PR #10, opened 2025-07-12
Init CPU

C0Florent:
First PR for the CPU!

I created a new crate where all CPU stuff is going to be implemented; defined a structure for registers; and implemented the most basic instruction to compute 1+1=2.

Minimal unit test ensures that 1 + 1 indeed equals 2.

Closes https://github.com/EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier/issues/7

### DIFF
--- a/cpu/src/registers.rs
+++ b/cpu/src/registers.rs
@@ -51,10 +51,23 @@ pub struct RegisterP {
     /// IRQ Disable flag
     pub I: bool,
 
-    /// Memory/Accumulator flag
+    /// Memory/Accumulator flag: Controls the width of the A register
+    ///
+    /// When set to 0, the accumulator register (A) is 16 bits wide \
+    /// When set to 1, the accumulator register (A) is 8 bits wide
+    ///
+    /// Always set to 1 in emulation mode, reset to 1 when switching to native mode
     pub M: bool,
 
-    /// Index flag
+    /// Index flag: Controls the width of the index registers X and Y
+    ///
+    /// When set to 0, both X and Y are 16 bits wide \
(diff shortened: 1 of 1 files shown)


ClementBaziret (dismissed):
LGTM

Maybe we could merge #11 before merging this PR ?

cpu/src/cpu.rs
```diff
@@ -0,0 +1,46 @@
...
+    ///
+    /// Flags set:
+    /// - `Z`: whether the result is zero
+    /// - `N`: whether the result is negative (if it were interpreted as signed)
+    ///
+    /// Returns the number of CPU cycles of execution
+    pub fn inx(&mut self) -> i32 {
+        self.registers.X = self.registers.X.wrapping_add(1);
+        self.registers.P.Z = self.registers.X == 0;
+        self.registers.P.Z = self.registers.X > 0x7fff;
```
Tholarr, line 26: What does this line do? I don't understand why is `Z` set again here 😅

C0Florent: Oh, it's meant to assign the N (negative) flag, not Z (zero)

As I pointed out in the doc comment right above:
> - `N`: whether the result is negative (if it were interpreted as signed)
So we're checking if the result (the X register) is greater than the maximum **signed** 16-bit integer (0x7fff), in which case it means that if the result would be read as a signed number, then it would be negative, so would should set the N bit to 1

Tholarr: :saluting_face:

Tholarr (approved):
All good :saluting_face:

PR #13, opened 2025-07-16
:sparkles: PPU Scanline rendering and basic data handling

Tholarr:
This MR introduces the foundation of the `PPU`, allowing an image to be **loaded**, **split into tiles**, and written into a temporary `VRAM`.

Using a basic loop, the `PPU` can **fetch tiles** from `VRAM` and **render** them to the screen via a **scanline-based rendering system**.

📌 Currently, the rendering is **black and white only**, as color support has not yet been implemented.

---

:warning: Please note that some components are still in early stages:
- **Tile** positioning is not yet handled. In the future, tiles will need to be associated with metadata (position, layer, etc.) like objects.
- **Color** support will require a dedicated memory system and more complex data management.

These limitations are covered by other issues and will be addressed in upcoming MRs :saluting_face: 

---
_Tileset used :_
tileset

_Result :_
Screenshot_2025-07-16_02-57-26

---
Also, CI has been fixed by:

- Expanding the coverage scope to all crates via the root `cargo.toml` file
- Excluding files in `tests/` directories and all `main.rs` files from coverage checks

---
Closes https://github.com/EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier/issues/5
Closes https://github.com/EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier/issues/18
Closes https://github.com/EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier/issues/19

### DIFF
--- a/ppu/src/tile.rs
+++ b/ppu/src/tile.rs
@@ -56,7 +56,7 @@ pub fn load_tiles_into_vram(ppu: &mut PPU, tiles: &Vec<Vec<Rgba<u8>>>) {
                 let pixel = &tile[y * TILE_SIZE + x];
 
                 // tmp 4 bits color saving => gonna be fixed when CGRAM ok (CGRAM not ok for PR #13]
-                let value = pixel[0] / 16;
+                let value = pixel[0] >> 4;
 
                 let addr = base_addr + y * TILE_SIZE + x;
                 ppu.write_vram(addr, value);
@@ -74,7 +74,7 @@ pub fn get_tile_from_vram(ppu: &PPU, tile_index: usize) -> Vec<u32> {
             let addr = base_addr + y * TILE_SIZE + x;
             let value = ppu.read_vram(addr);
 
-            let color = (value as u32) * 16;
+            let color = (value as u32) << 4;
             let argb = (0xFF << 24) | (color << 16) | (color << 8) | color;
             tile_pixels.push(argb);
         }


ppu/.gitignore
C0Florent, line 1: I assume you moved all your files into a ppu subfolder, including some which should have stayed in the repository root, such as this gitignore I would assume

ppu/README.md
C0Florent, line 1: should stay at repo root too?

ppu/.gitignore
Tholarr: Ohh yes mb, i wasn't thinking right at this time of the night :')

ppu/src/tile.rs
```diff
@@ -0,0 +1,84 @@
...
+use crate::ppu::PPU;
+use crate::utils::{TILE_SIZE};
+
+pub fn load_and_split_image(path: &str) -> (Vec<Vec<Rgba<u8>>>, usize) {
+    let img = image::open(path).expect("[ERR::ImageLoad] Failed to load image file.");
+    let img = img.to_rgba8();
+    let width = img.width() as usize;
+    let height = img.height() as usize;
+
+    let tile_size = TILE_SIZE as usize;
```
C0Florent, line 11:
```suggestion
    let tile_size = TILE_SIZE;
```

ppu/src/tile.rs
```diff
@@ -0,0 +1,84 @@
+use image::{Rgba};
+use crate::ppu::PPU;
+use crate::utils::{TILE_SIZE};
+
+pub fn load_and_split_image(path: &str) -> (Vec<Vec<Rgba<u8>>>, usize) {
```
C0Florent, line 5: We do agree that at some point this function will not be used anymore (at least as is)?
Of course we're not going to read assets from files, but rather from the ROM or some other place I guess

ppu/src/tile.rs
```diff
@@ -0,0 +1,84 @@
...
+            );
+            continue;
+        }
+
+        for y in 0..TILE_SIZE {
+            for x in 0..TILE_SIZE {
+                let pixel = &tile[y * TILE_SIZE + x];
+
+                // tmp 4 bits color saving => gonna be fixed when CGRAM ok (CGRAM not ok for PR #13]
+                let value = pixel[0] / 16;
```
C0Florent, line 59: Could write as a right shift by 4 instead of division by 16 (`x / 16` = `x >> 4`), more clearly expresses that we're keeping only the 4 upper bits perhaps?

ppu/src/tile.rs
```diff
@@ -0,0 +1,84 @@
...
+pub fn get_tile_from_vram(ppu: &PPU, tile_index: usize) -> Vec<u32> {
+    let mut tile_pixels = Vec::new();
+    let base_addr = tile_index * TILE_SIZE * TILE_SIZE;
+
+    for y in 0..TILE_SIZE {
+        for x in 0..TILE_SIZE {
+            let addr = base_addr + y * TILE_SIZE + x;
+            let value = ppu.read_vram(addr);
+
+            let color = (value as u32) * 16;
```
C0Florent, line 77: ... and then left shift by 4 to do the same as above? (`value as u32 << 4`)

ppu/src/utils.rs
```diff
@@ -0,0 +1,56 @@
...
+    )
+    .expect("[ERR::WindowInit] Unable to create display context.")
+}
+
+#[ignore] // Requires a graphical environment
+pub fn update_window(window: &mut Window, framebuffer: &Vec<u32>) {
+    window
+        .update_with_buffer(framebuffer, WIDTH, HEIGHT)
+        .expect("[ERR::Render] Framebuffer refused to cooperate.");
+}
```
C0Florent, line 34: Potentially move to main.rs if they are only used there for now?

ppu/tests/ppu_render.rs
```diff
@@ -0,0 +1,97 @@
...
+#[test] // Should create a window and update it with a pink framebuffer (because why not)
+#[ignore] // Requires a graphical environment
+fn test_create_window_and_update() {
+    let mut window = create_window();
+    assert!(window.is_open());
+
+    let framebuffer = vec![0xFF00FFu32; WIDTH * HEIGHT];
+
+    update_window(&mut window, &framebuffer);
+}
```
C0Florent, line 97: Perhaps remove this test if it can't be run

C0Florent (changes requested):
Seems good overall, I just left comments about some details we might want to tweak before merging

.github/workflows/actions.yaml
```diff
@@ -20,9 +20,11 @@ jobs:
     steps:
       - name: Checkout fetch
         uses: actions/checkout@v4.2.2
+      - name: Install g++
+        run: apt-get update && apt-get install -y g++
```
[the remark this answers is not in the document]
Tholarr: The minifb crate includes C++ code (scalar.cpp), so it needs g++

ppu/src/tile.rs
Tholarr: Yes, for the moment this function is only used to demonstrate that everything else is working, it's what provides me a test image since the memory isn't yet managed correctly and the components aren't connected

.github/workflows/actions.yaml
```diff
@@ -34,16 +36,16 @@ jobs:
...
       - name: Install openssl dev lib (necessary for cargo-tarpaulin)
         run: sudo apt install libssl-dev pkg-config -y
       - name: Install cargo tarpaulin
         timeout-minutes: 5
         run: cargo install cargo-tarpaulin
       - name: Check for failing unit test and coverage < COVERAGE_MINIMUM_PERCENTAGE
         timeout-minutes: 2
-        run: cargo tarpaulin --fail-under $COVERAGE_MINIMUM_PERCENTAGE
-      
+        run: cargo tarpaulin --workspace --fail-under $COVERAGE_MINIMUM_PERCENTAGE --exclude-files */tests/* */src/main.rs */main.rs
```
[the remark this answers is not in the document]
C0Florent: I couldn't resist checking for myself, and apparently it would be possible (although not in Cargo.toml)

The feature is documented [here](https://github.com/xd009642/tarpaulin/blob/HEAD/tarpaulin.toml) (link to actual paragraph)

Example config file (does include `fail-under` and `exclude-files`): https://github.com/xd009642/tarpaulin/blob/HEAD/tarpaulin.toml

ppu/src/utils.rs
Tholarr: I agree, but isn't it cleaner to separate tests from code?

ppu/tests/ppu_render.rs
Tholarr: :saluting_face:

ppu/src/utils.rs
C0Florent: Yes, but these two functions aren't tests? In this case you consider main to be the tests?

My point is actually that these functions are essentially just wrapping the dependency with some specific parameters which fit our current use case, but will be probably heavily reworked when we eventually link up the components together. I think this is why I see it better as something in the main.rs file which we know will be rewritten (and eventually removed?)

.github/workflows/actions.yaml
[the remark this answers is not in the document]
Tholarr: It works :)

ppu/src/tile.rs
Tholarr: :saluting_face:

tibaudlemaout (approved):
Seems good, as for most of the CI, i'm trusting Florent on this one

PR #18, opened 2025-09-13
Resumable CPU

C0Florent:
This PR adds a more advanced API for the CPU which should handle constraints imposed by other components for a good while.

I'll try to state what this is about:  
Previously, CPU instructions were standalone methods in the CPU struct, which means they had to be executed entirely, giving no way to any code managing the CPU to execute CPU cycles one by one. This what this PR is mostly about: giving a clean API to execute CPU cycles one by one.

Doing so also fulfills other constraints, such as returning the "type" of each executed cycle. This will be necessary for proper timing in the long run, because different types of cycles (whether the CPU accesses RAM or simply touches internal registers) may take varying amount of time to execute.

Implementing a per-cycle API forces us to add more fields to the CPU struct to "remember" what instruction we are executing and how many cycles of this instruction have already been executed, in contrary to previously when we were always executing entire instructions at a time.

Something deeply linked to implementing a per-cycle API is creating a to "automatically" read instructions. Previously, we were executing "known" instructions, with a dedicated public method for each instruction. This PR completely changes this by properly implementing *opcode fetches*: when you ask the CPU to execute a cycle, and its previous instruction is done, it *fetches* from RAM the *opcode* of the next instruction to execute.

This means that now, instructions "automatically" "chain", since calling a single `cycle` method several times will automatically fetch the next instruction after finishing the execution of the previous.

---

To have a clearer view of what this PR changes, feel free to review the change commit by commit, I've tried to split things up properly.


cpu/src/cpu.rs
```diff
@@ -1,31 +1,124 @@
...
 }
 
 impl CPU {
     pub fn new(registers: Registers) -> Self {
-        Self { registers }
+        Self {
+            registers,
+            addr_bus: SnesAddress::default(),
+            data_bus: 0,
+            IR: INSTRUCTIONS[0xea].unwrap(), // By default hold a NOP (no-operation) to do nothing
```
ClementBaziret, line 56: Is this the behavior of the snes cpu or is it just you putting a no-op in IR ?

cpu/src/cpu.rs
```diff
@@ -1,31 +1,124 @@
...
-    ///
-    /// Returns the number of CPU cycles of execution
-    pub fn inx(&mut self) -> i32 {
+    pub(crate) fn inx(&mut self) {
         self.registers.X = self.registers.X.wrapping_add(1);
         self.registers.P.Z = self.registers.X == 0;
         self.registers.P.N = self.registers.X > 0x7fff;
-
-        2
     }
```
ClementBaziret, line 122: In the future I think it would be better to put the instruction definition in another file(s), do you want to let the only one defined for now in `cpu.rs` or should we put it already in another file ?

ClementBaziret (commented):
LGTM, just some questions

tibaudlemaout (dismissed):
All of it seems good to me, but good luck filling up the instructions.

cpu/src/cpu.rs
```diff
@@ -1,31 +1,124 @@
...
+    /// Public getter to the address bus, needs to be read by the
+    /// code managing the CPU for RAM I/O
+    pub fn addr_bus(&self) -> &SnesAddress {
+        &self.addr_bus
+    }
+
+    /// Execute a single CPU cycle.
+    ///
+    /// This function is the core part of the public API to this struct.
+    /// TODO: Add more doc explaining simple use of this method
```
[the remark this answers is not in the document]
C0Florent: Yes, I can, and I probably should, you're right

cpu/src/cpu.rs
C0Florent: Indeed having all 256 opcodes implemented as standalone methods in this file will get unreadable quite soon. Having them all in a separate file would be better, although I don't know how well that would go either.

I think I'll keep adding a few here before I can envision what would be a better solution in the long run, for now I don't really know since I haven't implemented many instructions yet.

Honestly I'm thinking of creating some sort of meta-language (a made up language that will be translated to Rust) to implement all instructions in a concise manner, because most instructions repeat the same kind of operations (like reading a value from an address and then incrementing the read address, combining two bytes into a u16, adding an execution cycle when a memory access crosses a page boundary, incrementing things with `wrapping_add`), and I feel like even with nice helpers it would get hard to read given the number of instructions to implement.
I think I'd want to implement this meta-language with Rust proc macros, which allow defining arbitrary syntax within Rust code that will be expanded at compile-time.

cpu/src/instr_tab.rs
```diff
@@ -0,0 +1,286 @@
...
+/// the [`Internal`] variant holds a function pointer to the rust code to run.
+pub(crate) enum InstructionCycle {
+    Read,
+    Write,
+    Internal(fn(&mut CPU) -> ()),
+}
+
+/// Type representing the definition of a CPU instruction
+///
+/// Conceptually, an instruction is simple a list of actions
```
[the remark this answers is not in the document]
C0Florent: yes, that would make more sense

cpu/src/cpu.rs
ClementBaziret: ok 🫡

cpu/src/cpu.rs
ClementBaziret: Yeah this could be cool but don't start doing something that will take you weeks before even starting to implement the different operations.
You can definitely try if you want but maybe creating some helpers function and do everything "by hand" is the fastest way idk.

C0Florent: That's why I plan to implement a good number of instructions "by hand" first, don't worry I'll try my best to avoid going too deep in "simplifying" things which don't need to be simplified

ClementBaziret: ok ok 🫡

cpu/src/cpu.rs
```diff
@@ -43,4 +184,36 @@ mod test {
         cpu.inx();
         assert_eq!(cpu.regs().X, 2);
     }
+
+    #[test]
+    fn test_1_plus_1_is_2_cycle_api() {
```
Tholarr, line 189: Do we want to organize the tests in the file they are testing?

C0Florent: I'm going to answer this the same way as [this comment](https://github.com/r-snes/r-snes/pull/18/files/14e4b56687ebefd99f2bfa2866357f024e976f0c#r2365417226): yes there will be some (probably many) things to move *in the future*, but I think I would need to have more stuff done in the first place before I can see how things should be organised.

And in this particular case, this test is testing things implemented in several files (including this one), so I don't know for now what would be a better place for it.

Tholarr: alright thx :saluting_face:

Tholarr (approved):
lgtm

tibaudlemaout (approved):
Looking good :+1:

PR #21, opened 2025-10-04
U16Split

C0Florent:
Simple PR which adds a utility in our `common` crate: `u16_split`

It allows splitting (immutable/mutable) references to a `u16` into two references to its individual bytes (as references to `u8`s).

The implementation required to use `unsafe` to make use of raw pointers to create references to `u8`. I believe I respected the constraints of `unsafe` which makes these functions safe to use.

---

I also tested tested whether the compiler was able to optimise the use of these functions when accessing a single byte of a `u16`, and when enabling enough optimisations (from `-C opt-level=2` onwards IIRC): the two following functions compile to the same assembly code (so the compiler *is* indeed able to optimise out the abstractions properly).
```
#[inline(never)]
pub fn set_hi1(mut x: u16) -> u16 {
    *x.hi_mut() = 0xfe;
    x
}
#[inline(never)]
pub fn set_hi2(x: u16) -> u16 {
    (x & 0x00ff) | 0xfe00
}
```
(i had to use `#[inline(never)]` because from opt level 2 the compiler decided to always inline the functions and stopped producing assembly code for them when compiling them by themselves)
(actually even better than producing the same assembly code, the compiler made one function an alias of the other (from what I understand of the following assembly)):
```s
	.intel_syntax noprefix
	.file	"test1.8159662b3423405c-cgu.0"
	.section	.text._ZN5test17set_hi117had2caf20830d439cE,"ax",@progbits
	.globl	_ZN5test17set_hi117had2caf20830d439cE
	.p2align	4
	.type	_ZN5test17set_hi117had2caf20830d439cE,@function
_ZN5test17set_hi117had2caf20830d439cE:
	.cfi_startproc
	movzx	eax, dil
	or	eax, 65024
	ret
.Lfunc_end0:
	.size	_ZN5test17set_hi117had2caf20830d439cE, .Lfunc_end0-_ZN5test17set_hi117had2caf20830d439cE
	.cfi_endproc

	.globl	_ZN5test17set_hi217h305aee47e557012eE
	.type	_ZN5test17set_hi217h305aee47e557012eE,@function
.set _ZN5test17set_hi217h305aee47e557012eE, _ZN5test17set_hi117had2caf20830d439cE
	.ident	"rustc version 1.89.0 (29483883e 2025-08-04) (built from a source tarball)"
	.section	".note.GNU-stack","",@progbits
```

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -1,1 +1,2 @@
 pub mod snes_address;
+pub mod u16_split;
--- a/common/src/u16_split.rs
+++ b/common/src/u16_split.rs
@@ -1,0 +1,87 @@
+pub trait U16Split {
+    fn split<'a>(&'a self) -> SplitU16<'a>;
+    fn split_mut<'a>(&'a mut self) -> SplitU16Mut<'a>;
+
+    fn lo<'a>(&'a self) -> &'a u8 {
+        self.split().lo
+    }
+    fn hi<'a>(&'a self) -> &'a u8 {
+        self.split().hi
+    }
+
+    fn lo_mut<'a>(&'a mut self) -> &'a mut u8 {
+        self.split_mut().lo
+    }
+    fn hi_mut<'a>(&'a mut self) -> &'a mut u8 {
+        self.split_mut().hi
+    }
+}
+
+pub struct SplitU16<'a> {
+    pub lo: &'a u8,
+    pub hi: &'a u8,
+}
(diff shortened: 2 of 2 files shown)


ClementBaziret (approved):
LGTM 🫡

tibaudlemaout (approved):
Looks goo to me 👍

ClementBaziret (approved):
LGTM 🫡

PR #23, opened 2025-10-04
:sparkles: Init CPU meta-language

C0Florent:
"Simple" PR for the issue mentioned below.
This PR implements a Rust procedural macro that will help me write
CPU instructions in a concise way. It's very likely that I will
change many things in the code added by this PR in the near future,
but I feel like it's good to let you review the early version which
already works and can be unit-tested properly.

Feel free to ask any question about weird things I do with the token
stream manipulation functions, a lot of this can be unintuitive.

closes EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier#23

### DIFF
--- a/cpu/instr_metalang_procmacro/src/parser.rs
+++ b/cpu/instr_metalang_procmacro/src/parser.rs
@@ -1,0 +1,116 @@
+use pm2::{Ident, TokenStream, TokenTree};
+use proc_macro2 as pm2;
+
+pub(crate) enum MetaInstruction {
+    EndCycle(Ident),
+    EndInstr(Ident),
+}
+
+impl MetaInstruction {
+    fn is_end_instr(&self) -> bool {
+        match self {
+            Self::EndInstr(_) => true,
+            _ => false,
+        }
+    }
+}
+
+impl MetaInstruction {
+    fn try_from<I: IntoIterator<Item = TokenTree>>(value: I) -> Result<Self, &'static str> {
+        let mut it = value.into_iter();
+
+        let Some(TokenTree::Ident(meta_kw)) = it.next() else {
+            Err("Expecting a meta-keyword")?
+        };
+        let ret = match meta_kw.to_string().as_str() {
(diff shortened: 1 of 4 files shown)


cpu/instr_metalang_procmacro/src/parser.rs
```diff
@@ -0,0 +1,116 @@
...
+impl MetaInstruction {
+    fn is_end_instr(&self) -> bool {
+        match self {
+            Self::EndInstr(_) => true,
+            _ => false,
+        }
+    }
+}
+
+impl MetaInstruction {
```
Tholarr, line 21: Why having two separate `impl MetaInstruction` blocks ?

cpu/instr_metalang_procmacro/src/parser.rs
```diff
@@ -0,0 +1,116 @@
...
+        let mut ret = Instr::new(name);
+        loop {
+            let it = it.by_ref();
+            let cycle_body = it.take_while(|token| token.to_string() != "meta");
+            let body_tok_stream = TokenStream::from_iter(cycle_body);
+            let meta_instr = MetaInstruction::try_from(it.take_while(|token| {
+                let TokenTree::Punct(p) = token else {
+                    return true;
+                };
+                return p.as_char() != ';';
```
Tholarr, line 94: What happens if there is a `;` in a different position than the one specified here?

cpu/instr_metalang_procmacro/src/parser.rs
C0Florent: I guess while developing hoping to get the second block to be a `impl TryFrom<I: IntoIterator<Item = TokenTree>> for MetaInstruction`, but the compiler didn't agree with that, so I had to replace it to a simple `impl` block, without merging it with the one above.

I might try to get it working again later, so I'd rather keep it separate for now

tibaudlemaout (approved):
It seems good to me, documentation is great as well

PR #49, opened 2026-02-07
:sparkles: Main emulator program

ClementBaziret:
## 📝 Description

This pull requests creates a first draft of the main program. The main program is the one that is displaying a window to the user and managing the different emulator components.
It's the main program that will for example call the cpu `cycle` function when needed.

For now there isn't anything related to the other components, it will come later in other pull requests.

## 🔍 Related Issues

Closes https://github.com/EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier/issues/54


src/main.rs
```diff
@@ -1,19 +1,64 @@
...
+        // Emulation update if emulator exists and if master_cycle duration treshold is crossed
+        match rsnes_app {
+            Some(ref mut app) => {
+                master_cycle_accum += delta;
+
+                // In the future, check if this 'while' doesn't restrain the program execution too much
+                while master_cycle_accum >= Rsnes::MASTER_CYCLE_DURATION {
+                    master_cycle_accum -= Rsnes::MASTER_CYCLE_DURATION;
+                    app.update();
+                }
```
C0Florent: Actually I think this is perfectly fine: we're potentially catching up missed frames in case the previous loop took too long for some reason

src/gui.rs
```diff
@@ -0,0 +1,147 @@
...
+        self.clear(30, 30, 35);
+        match self.handle_events(rsnes_app) {
+            Some(Event::Quit { .. }) => return Err("User quit window".to_string()),
+            _ => {}
+        }
+        let _ = self.draw_framebuffer(); // TODO: Handle error properly
+        self.present();
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
ClementBaziret: I believe it is fixed with the rework of the `Gui` struct `impl`

src/rsnes.rs
```diff
@@ -0,0 +1,81 @@
...
+use ppu::ppu::PPU;
+
+use std::error::Error;
+use std::path::Path;
+use std::path::PathBuf;
+
+pub struct RSnes {
+    pub rom_path: PathBuf,
+    pub bus: Box<Bus>,
+    pub cpu: Box<CPU>,
```
C0Florent: Don't need to box the CPU, that's for sure. Perhaps you don't need to box all other things, but I'm not as certain

C0Florent (commented):
Apparently I missed some details in my last review

C0Florent (approved):

tibaudlemaout (approved):
LGTM

PR #52, opened 2026-03-07
:sparkles: Implement transfer instructions

C0Florent:
## 📝 Description

This PR implements transfer instructions, which simply *transfer* the value of one register to another. Some have special cases for variable-width registers (A, X and Y), but nothing too complicated.

## 🔍 Related Issues

None


ClementBaziret (dismissed):
LGTM

cpu/src/instrs/transfers.rs
```diff
@@ -0,0 +1,188 @@
...
+        } else {
+            cpu.registers.S = cpu.registers.DUP_src;
+        }
+
+        meta END_CYCLE Internal;
+    });
+}
+
+// duplicate over variable width registers (as destination --
+// since the size of the transfer comes from the size of the source)
```
Tholarr: This comment states that the size of the transfer comes from the source registry, but it contradicts the comment at the top of the file.
Line 5: `//! transfer is determined by the size of the destination register`

 Which one is correct?

C0Florent: hmmm

C0Florent: Well, nice catch, as stated in the module documentation (top of file), and as earlier in the same comment, the size of the transfer does depend on the size of the destination, not of the source.
At least I got it right in the implementation

Tholarr (approved):
lgtm 👍

tibaudlemaout (approved):

PR #53, opened 2026-03-07
:sparkles: Implement CPU miscellaneous instructions

C0Florent:
## 📝 Description

Implements various instructions, nearly everything still missing the CPU (only excluding stack instructions and interrupt-related instructions)

## 🔍 Related Issues

None

PR #54, opened 2026-03-07
:sparkles: Implement Stack instructions

C0Florent:
## 📝 Description

This PR implements all "stack" instructions: instructions which pull and push things to the stack. Some minor corrections will have to be made later with things that use `PUSHN`, but hopefully we should see fails from the snes-test ROM to remind us before it becomes meaningful, it's really a rare edge case.

*Note*: Has to be merged **after** #52 and #53 (conflicts due to lines added at similar places which i don't want to bother resolving)

## 🔍 Related Issues

None


Tholarr (dismissed):
lgtm 👍

ClementBaziret (approved):
As always the review of new opcodes is a bit silly because we read and we're like yeah seems good but we don't know what it's supposed to do but well LGTM.

Tholarr (approved):
👍

PR #55, opened 2026-03-28
:sparkles: Plugin init

C0Florent:
## 📝 Description

Initialises the base architecture for plugin support, all in a new `plugins` crate.
Some code is copied over from the PoC on rboy we did last year (the Permission and StrictPartialOrd stuff).

For now, we only:
- load and evaluate a lua file with piccolo (but don't do anything with the returned value)
- ask the user for permissions in a GUI window with egui (but just a single checkbox)
- react depending on whether user granted perms (by printing in the terminal)

Following PRs should be able to improve each of these 3 main axes individually (plugin table parsing, more detailed GUI, and actually loading things in the lua VM depending on granted permissions)

## Demo

https://github.com/user-attachments/assets/92392bad-0807-4d16-bd84-b41b7b4d9e5a

## 🔍 Related Issues

none (oops, i'll make some for the individual parts for next PRs)


tibaudlemaout (approved):
This is great, plus the plan.txt helps imagining how the plugins can take shape 👍

plugins/plan.txt
```diff
@@ -0,0 +1,114 @@
...
+```
+return {
+    permissions = "all",
+    actions = {
+        default = function(rsnes)
+            rsnes.simple_dialog("hello world")
+        end
+    }
+}
+```
```
ClementBaziret, line 16: Do you think it could be a good idea to remove the `return` from the plugin files ? This would avoid users to have to write the return even though it will most likely always need to be here and we can instead add it on our side in the plugin reading.

ClementBaziret (approved):

plugins/plan.txt
C0Florent: I'm really unsure, as basically we load the entire plugin file as a function body, so it kinda has to return something.

We could definitely prepend a `return ` at the start of the file, but then it would break if the user wants to define some variables before their `return` statement, for example:

```lua
local requested_perms = "all"
return {
    permissions = requested_perms,
    actions = {
        default = function(rsnes)
            rsnes.simple_dialog("hello world")
        end
    }
}
```
(this particular example doesn't make a lot of sense but you can probably imagine cases where it does)

ClementBaziret: Yep ok seems okay to me then.

PR #56, opened 2026-04-04
BRR Decoding Fixes + APU Structural Refactor

tibaudlemaout:
## 📝 Description

What changed for the various APU component:

BRR decoder: 

- Has been added
- Fixed header bit layout
- 15-bit output clamp, end/loop flag handling 
- All of these caused incorrect or garbage audio output.

DSP: 

- Corrected per-voice register layout
- ADSR1 bit layout
- ENVX/OUTX/ENDX register updates
- Master volume wired through to the mixer.

APU:

- Removed unused DSP field 
- Fixed render_audio
- Corrected DSP tick rate to 32 CPU cycles per sample
- Added cpu.reset() on init.

Memory:

- Implemented the full $F0–$FF I/O map including the $F2/$F3 DSP port protocol
- $F4–$F7 bidirectional comm ports,
- Timer divisors/counters
- CONTROL register

CPU:

- Corrected &mut Memory to &Memory on all read-only instructions
- Address fetch logic using read_immediate16.

If you wish to test the sounds created by main you will need to install ffmpeg version 6.1.1
⚠️ The sounds are very harsh to the ears

## 🔍 Related Issues

Closes EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier#37


apu/src/apu.rs
```diff
@@ -1,38 +1,88 @@
...
+                self.dsp_cycles = 0;
+
+                // dsp.step() needs to read BRR sample data from APU RAM, but
+                // Memory owns the Dsp, so we cannot pass &self.memory while
+                // also holding &mut self.memory.dsp.  We resolve this by
+                // building a read-only RAM view.  This will be eliminated when
+                // dsp.step() is refactored to take a &[u8] RAM slice directly
+                // (see issue tracker — "borrow conflict" item).
+                let ram_snapshot = self.memory.ram;
+                self.memory.dsp.step_with_ram(&ram_snapshot);
```
C0Florent: you can just `self.memory.dsp.step_with_ram(&self.memory.ram);`

apu/src/dsp.rs
```diff
@@ -1,115 +1,275 @@
...
+const ENVELOPE_RATE_TABLE: [u16; 32] = [
+    0,    // 0: never (infinite)
+    2048, 1536, 1280, 1024, 768,
+    640,  512,  384,  320,  256,
+    192,  160,  128,  96,   80,
+    64,   48,   40,   32,   24,
+    20,   16,   12,   10,   8,
+    6,    5,    4,    3,    2,
+    1,    // 31: every tick
+];
```
C0Florent: Why not have the big descriptive comment as part of the doc comment below?

Perhaps leave the current doc comment at the start and add the expanded description as a second paragraph to it, kind of like the following:
```
/// Ticks between envelope updates for each rate index (0–31).
/// Rate 0 = never update (infinite). Rate 31 = update every tick.
///
/// The real DSP uses a 32-entry lookup table to determine how
/// many ticks pass between each envelope step. Index 0 means
/// "never" (infinite hold). All other values are tick counts.
```
I think it should also be rephrased to not mention twice that index 0 is "never"

the point of having doc comments is that they will be picked up when generating documentation with `cargo doc`, so i think all relevant info can be included in them.

apu/src/dsp.rs
```diff
@@ -1,115 +1,275 @@
...
+    /// determined by the rate table. Each phase has its own rate source.
     pub fn update_envelope(&mut self) {
         match self.envelope_phase {
+            // -------------------------------------------------
+            // ATTACK: rise linearly from 0 to 0x7FF.
+            // Rate index comes from the 4-bit attack_rate field.
+            // Special case: attack_rate == 15 uses a fixed +1024 step
+            // for a near-instant attack.
+            // -------------------------------------------------
             EnvelopePhase::Attack => {
```
C0Florent: Could this comment block not be placed on the enum variant instead?

like
```rs
/// enum doc
enum Something {
    /// documentation on enum variants!
    VariantA,

    /// each variant has its description
    VariantB,
}
```

same reason as previously, by placing relevant info in doc comments, they can be picked up by `cargo doc`

(same comment for the other match arms below)

apu/src/dsp.rs
```diff
@@ -1,115 +1,275 @@
...
+                    let rate_idx = (self.attack_rate * 2 + 1) as usize;
+                    let period = ENVELOPE_RATE_TABLE[rate_idx.min(31)];
+                    if !self.tick_due(period) {
+                        return;
+                    }
+                    self.envelope_level = (self.envelope_level + 32).min(0x7FF);
+                }
 
                 if self.envelope_level >= 0x7FF {
                     self.envelope_level = 0x7FF;
```
C0Florent: If i understood well, I think the `self.envelope_level = 0x7FF` will never do anything as it can only ever be `0x7FF` thanks to the `.min(0x7FF)` you use above, so it should be safe to remove, and the `>= 0x7FF` should be replaceable with `== 0x7FF`

if you make this change i think it would be best to at least add a comment to make it explicit that this relies on the `.min(0x7FF)` above

apu/src/dsp.rs
C0Florent, line 1: I think it would help in readability to have this file split on multiple files
perhaps make a `dsp` folder, with separate files for ADSR, BRR, Voices, and the DSP itself

apu/src/dsp.rs
```diff
@@ -118,225 +278,717 @@ impl Default for Voice {
...
-            current_addr: 0,
-            frac: 0,
+            pitch_counter: 0,
             current_sample: 0,
-
             adsr: Adsr::default(),
+            brr: Brr::default(),
         }
     }
 }
```
C0Florent: just realising now: if this is your implementation of `Default`, you might as well just `#[derive(Default)]`, which uses 0 for integers and calls `default()` to initialise other members.

apu/src/apu.rs
```diff
@@ -1,38 +1,80 @@
...
+            // Advance the full APU by one DSP period (32 CPU cycles = 1 sample).
+            self.step(DSP_CYCLES_PER_SAMPLE);
+
+            // Collect the stereo output from the DSP.
+            let (l, r) = self.memory.dsp.render_audio_single();
+            buf.push(l);
+            buf.push(r);
+        }
+
+        buf
```
C0Florent: If semantically the output vector is always made of pairs, why not make it explicit as a `Vec<(i16, i16)>`

So you would change the capacity to only be `num_samples` (since now all elements have two values), and you could replace the two `buf.push` by a single `buf.push((l, r))`, or directly `buf.push(self.memory.dsp.render_audio_single());`

apu/src/dsp.rs
```diff
@@ -1,115 +1,273 @@
...
+            addr: 0,
+            nibble_idx: 0,
+            prev1: 0,
+            prev2: 0,
+            loop_addr: 0,
+            sample_buffer: [0i16; 16],
+            buffer_fill: 0,
         }
     }
 }
```
C0Florent: Could simply be a `#[derive(Default)]`

apu/src/dsp.rs
```diff
@@ -1,115 +1,273 @@
...
 
-/// Voice of the SNES APU DSP
+// ============================================================
+// VOICE
+// Represents one of the 8 independent audio channels.
+// ============================================================
+
+/// One voice (channel) of the SNES APU DSP.
```
C0Florent: I'm not sure about the non-doc comment, it doesn't really add anything?

apu/src/dsp.rs
```diff
@@ -118,225 +276,529 @@ impl Default for Voice {
...
-            // Sets the high 8 bits of the pitch value.
-            0x18..=0x1F => {
-                let voice_idx = index - 0x18;
-                let pitch = ((value as u16) << 8) | (self.voices[voice_idx].pitch & 0x00FF);
-                self.voices[voice_idx].pitch = pitch;
+        match (voice_num, reg_off) {
+            // ---- Per-voice registers (voices 0-7) ----
+
+            // +0: VOL(L) — signed left volume
+            (v, 0x0) if v < 8 => self.voices[v].left_vol = value as i8,
```
C0Florent: useless `if v < 8` (and same in other match arms below):
- `v` is `voice_num`
- `voice_num` is `idx >> 4`
- `idx` is `index & 0x7F`

so:
- `idx` is at most 127
- `voice_num` is at most 7
- v is at most 7

if you wanted to make it extra explicit, you could add an `assert!(voice_num < 7);` or a comment saying it, but i think it's irrelevant to check it in all match arms

apu/src/dsp.rs
```diff
@@ -118,225 +276,529 @@ impl Default for Voice {
...
-                    v.current_addr = v.sample_start;
-                    v.frac = 0;
-                    v.adsr.envelope_phase = EnvelopePhase::Attack;
-                    v.adsr.envelope_level = 0;
-                }
+            // +3: PITCH high byte (only bits 5-0 = pitch bits 13-8)
+            (v, 0x3) if v < 8 => {
+                let p = &mut self.voices[v].pitch;
+                *p = (*p & 0x00FF) | ((value as u16 & 0x3F) << 8);
             }
```
C0Florent: Could replace with `*p.hi_mut() = value & 0x3F` if i undestand that correctly

apu/src/dsp.rs
```diff
@@ -118,225 +276,529 @@ impl Default for Voice {
...
 
-            // 0x38..=0x3F: Sample End (low byte)
-            // Sets the low 8 bits of the sample end address.
-            0x38..=0x3F => {
-                let voice_idx = index - 0x38;
-                self.voices[voice_idx].sample_end =
-                    (self.voices[voice_idx].sample_end & 0xFF00) | value as u16;
+            // +7: GAIN — TODO: implement GAIN mode
+            (v, 0x7) if v < 8 => { /* GAIN mode not yet implemented */ }
```
C0Florent:
```suggestion
            // +7: GAIN — TODO: implement GAIN mode
            (v, 0x7) if v < 8 => todo!("GAIN mode"),
```

apu/src/dsp.rs
```diff
@@ -118,225 +276,529 @@ impl Default for Voice {
...
-                self.voices[voice_idx].sample_start =
-                    (self.voices[voice_idx].sample_start & 0xFF00) | value as u16;
+            // +6: ADSR2 = SSSRRRRR
+            //   bits 7-5: sustain level (0–7)
+            //   bits 4-0: sustain rate index (0–31)
+            (v, 0x6) if v < 8 => {
+                let adsr = &mut self.voices[v].adsr;
+                adsr.sustain_level = (value >> 5) & 0x07;
+                adsr.sustain_rate  =  value & 0x1F;
             }
```
C0Florent: Just a question about that: is there technically some duplicated info between `self.registers` and the voices' ADSR? (storing both the raw byte in `self.registers` and the "decoded" value in the ADSR)

apu/src/dsp.rs
```diff
@@ -118,225 +276,529 @@ impl Default for Voice {
...
+            // $1C: MVOLR — master right volume (signed)
+            (_, _) if idx == 0x1C => self.master_vol_right = value as i8,
+
+            // $5D: DIR — sample directory base page
+            (_, _) if idx == 0x5D => {
+                self.dir_base = value;
             }
 
-            _ => {} // Other registers (echo, gain, FIR, etc.) not implemented yet
+            _ => {}
```
C0Florent: I'm not 100% sure but it might be clearer to wrap all global registers in a single arm like ` _ => match idx { ... }` (to avoid the weird repetition of the `(_, _) if idx ==`

apu/src/dsp.rs
```diff
@@ -118,225 +276,529 @@ impl Default for Voice {
...
+    // Takes a plain &[u8] RAM slice so the caller can pass
+    // &memory.ram without also borrowing memory.dsp mutably.
+    // ----------------------------------------------------------
+
+    /// Advance the DSP by one output sample tick.
+    ///
+    /// `ram` is a direct slice of the 64 KB APU RAM.  The DSP only
+    /// reads from RAM (BRR sample data and the DIR table); it never
+    /// writes to it.
+    pub fn step(&mut self, ram: &[u8]) {
```
C0Florent: (about the entire function)

1. (easy) the loop could be replaced by a `for (i, v) in self.voices.iter_mut().enumerate()`. so `i` correspond to the current `v`, and `v` would be `&mut self.voices[i]`
2. (harder) could the loop body be extracted in a method of `Voice`? it could help readability but i'm not sure it's easily feasible

apu/src/memory.rs
```diff
@@ -1,24 +1,142 @@
...
+// $F4–$F7    CPUIO0–3  — bidirectional communication ports (CPU ↔ APU)
+// $F8–$F9    AUXRAM    — extra RAM bytes (readable/writable normally)
+// $FA–$FC    TnDIV     — timer 0/1/2 divisor (write-only)
+// $FD–$FF    TnOUT     — timer 0/1/2 counter output (read-only, clears on read)
+//
+// The direct-mapped range $F200–$F27F used by test code is kept alongside
+// the real port protocol so both can coexist during development.
+// ============================================================
+
 pub struct Memory {
```
C0Florent: Couldn't this just be a doc comment of the struct?

apu/src/memory.rs
```diff
@@ -1,24 +1,142 @@
...
+
+            // ---- Normal RAM ----
+            _ => self.ram[addr as usize],
+        }
+    }
+
+    /// Like read8 but takes &mut self so that reading $FD–$FF can clear
+    /// the timer output counters, matching real hardware behaviour.
+    /// The SPC700 CPU must call this variant when executing load instructions.
+    pub fn read8_mut(&mut self, addr: u16) -> u8 {
```
C0Florent, line 132: since this version should "always" be used instead of `read`, are there any cases where `read` would legitimately be called in the final project?

I'm trying to see if we could make it private or something like this, to avoid having code call `read` instead of `read_mut`, which could cause hard to spot inaccuracies in the long run

apu/src/memory.rs
```diff
@@ -28,18 +146,82 @@ impl Memory {
...
     }
 
     pub fn write16(&mut self, addr: u16, value: u16) {
-        let lo = (value & 0xFF) as u8;
-        let hi = (value >> 8) as u8;
-        self.write8(addr, lo);
-        self.write8(addr.wrapping_add(1), hi);
+        self.write8(addr,                      (value & 0xFF) as u8);
+        self.write8(addr.wrapping_add(1), (value >> 8) as u8);
+    }
```
C0Florent, line 199: Could use `*value.lo()` and `*value.hi()`

apu/src/apu.rs
tibaudlemaout: This has been fixed in the "removed duplicated code..." commit, it is indeed a lot cleaner, thanks.

apu/src/dsp.rs
tibaudlemaout: Yes, it's intentional. 
`self.registers` handles read-back via `$F3`, while the decoded fields make `step()` faster at 32 kHz. `write_reg` keeps both in sync.

apu/src/dsp.rs
tibaudlemaout: the second option has been implemented, most of the step logic now resides in voice.

apu/src/memory.rs
tibaudlemaout: read8 is still needed, read8_mut delegates to it for non-timer addresses and read16 uses it.

apu/src/dsp/brr.rs
```diff
@@ -0,0 +1,168 @@
...
+    ram: &[u8],
+    addr: u16,
+    prev1: &mut i16,
+    prev2: &mut i16,
+) -> ([i16; 16], bool, bool) {
+    let header = ram_read8(ram, addr);
+
+    let shift  = (header >> 4) & 0x0F;
+    let filter = (header >> 2) & 0x03;
+    let looop  = (header & 0x02) != 0;
```
Tholarr, line 147: an extra `o` slipped in

apu/src/dsp/mod.rs
```diff
@@ -0,0 +1,254 @@
+mod adsr;
+mod brr;
+mod voice;
+
+// Re-export everything tests and external code need
+pub use adsr::{Adsr, EnvelopePhase};
+use adsr::ENVELOPE_RATE_TABLE;
```
Tholarr, line 7: Is there a reason for importing `adsr` twice ?

apu/src/apu.rs
```diff
@@ -1,38 +1,78 @@
...
-    /// Generate audio samples (stub for now)
-    pub fn render_audio(&mut self, _num_samples: usize) -> Vec<i16> {
-        // TODO: pull from DSP’s audio buffer
-        vec![0; _num_samples]
+    /// Generate `num_samples` stereo output samples.
+    ///
+    /// Steps the APU internally for each sample so that CPU, timers, and DSP
+    /// all advance in lock-step.  Returns a `Vec` of `(left, right)` pairs.
+    pub fn render_audio(&mut self, num_samples: usize) -> Vec<(i16, i16)> {
+        let mut buf = Vec::with_capacity(num_samples);
```
Tholarr: Maybe a dumb question but isn't it supposed to be `buff` ?

Tholarr (changes requested):
LGTM, there are just some typos :)

apu/src/dsp/brr.rs
tibaudlemaout: The reason why is because loop is a keyword, therefore the extra o was to not name it something convoluted.

apu/src/dsp/mod.rs
tibaudlemaout: it was to avoid making all 3 of them public if not needed, the first 2 are needed but not the third, hence the missing pub on line 7.

apu/src/apu.rs
tibaudlemaout: it has been chaneged to buff.

apu/src/dsp/brr.rs
Tholarr: ohh alright mb

apu/src/dsp/mod.rs
Tholarr: alright 👍

apu/src/apu.rs
Tholarr: change has been made, all good 👍

apu/src/apu.rs
```diff
@@ -1,38 +1,78 @@
...
+// The DSP produces one output sample every 32 CPU cycles (32 kHz).
+// We count CPU cycles and only tick the DSP when this threshold is reached.
+const DSP_CYCLES_PER_SAMPLE: u32 = 32;
 
 pub struct Apu {
-    pub cpu: Spc700,
-    pub dsp: Dsp,
-    pub memory: Memory,
+    pub cpu:    Spc700,
+    pub memory: Memory,   // Memory::dsp is the *only* Dsp — there is no separate field
```
ClementBaziret: Is this comment relevant ? Why would there be another field ?
` // Memory::dsp is the *only* Dsp — there is no separate field`

apu/src/apu.rs
```diff
@@ -1,38 +1,78 @@
...
-    /// Step the APU forward by a given number of cycles
+    /// Step the APU forward by `cycles` CPU cycles.
+    ///
+    /// Each call ticks:
+    ///   - The SPC700 CPU  (every cycle)
+    ///   - The timers      (every cycle)
+    ///   - The DSP         (once every 32 cycles → 32 kHz)
+    ///
+    /// All DSP access goes through `self.memory.dsp`; there is no
+    /// separate Dsp field on Apu.
```
ClementBaziret: Same here, is this comment necessary too ?

apu/src/memory.rs
```diff
@@ -1,24 +1,138 @@
...
+    ///
+    /// The SNES main CPU writes to the APU side of these ports; the SPC700
+    /// reads them as inputs.  The SPC700 can also write them; the main CPU
+    /// reads those as inputs.  We store both directions separately so that
+    /// neither side clobbers the other's data.
+    ///
+    /// `port_in[n]`  — value written by the SNES CPU, read by SPC700 at $F4+n
+    /// `port_out[n]` — value written by SPC700 at $F4+n, read by SNES CPU
+    pub port_in:  [u8; 4],
+    pub port_out: [u8; 4],
```
ClementBaziret, line 55: Does the communication port with CPU really have 2 separate "channels" of communication for the 2 directions ?

apu/src/memory.rs
```diff
@@ -1,24 +1,138 @@
...
-        if (0xF200..=0xF27F).contains(&addr) {
-            self.dsp.read(addr)
-        } else {
-            self.ram[addr as usize]
+        match addr {
+            // ---- SPC700 I/O ports ($00F0–$00FF) ----
+
+            // $F0 TEST — write-only; reads return 0xFF on hardware but 0 is
+            // safe for emulation purposes since no game reads it.
+            0x00F0 => 0x00,
```
ClementBaziret: Why not returning `0xFF` if that's what the console does ?

apu/src/memory.rs
```diff
@@ -1,24 +1,138 @@
...
+    /// the timer output counters, matching real hardware behaviour.
+    /// The SPC700 CPU must call this variant when executing load instructions.
+    pub fn read8_mut(&mut self, addr: u16) -> u8 {
+        match addr {
+            0x00FD => { let v = self.timer_out[0]; self.timer_out[0] = 0; v }
+            0x00FE => { let v = self.timer_out[1]; self.timer_out[1] = 0; v }
+            0x00FF => { let v = self.timer_out[2]; self.timer_out[2] = 0; v }
+            _      => self.read8(addr),
         }
     }
```
ClementBaziret, line 139: I personally think that this function is kind of weird, in my opinion it would be better to merge the 2 read functions and make them take a mut ref by default. Because otherwise it is the role of other components to handle the read that reset the values since we'll have to call this specific function for it.

ClementBaziret: Nop sry, I misunderstood and did not really see the `self.read8`.

ClementBaziret (changes requested):
LGTM overall, just some questions for which I made comments

apu/src/apu.rs
tibaudlemaout: This comment was added during a commit getting rid of code duplication for ease of access to the DSP, it has been removed as it is indeed no longer relevant.

apu/src/apu.rs
tibaudlemaout: The comment has been modified to this: ` /// All DSP access goes through `self.memory.dsp`;` to still know how to access the DSP from the APU core without the mention of the previously mentioned redundant duplication.

apu/src/memory.rs
tibaudlemaout: I was using '0x00' as it was already implemented, which is not the case for `0xFF` but since it shouldn't have an impact regardless it has been changed to match and now returns `0xFF`.

apu/src/memory.rs
tibaudlemaout: Yes, ports $F4–$F7 go both ways but the two directions are independent. The main CPU writes to its side of the latch and the SPC700 reads from it, while the SPC700 writes to its side and the main CPU reads from that. 

port_in[n] is what the main CPU sends for the SPC700 to read at $F4+n, and port_out[n] is what the SPC700 writes for the main CPU to read.

apu/src/apu.rs
ClementBaziret: noice

apu/src/apu.rs
ClementBaziret: noice

apu/src/memory.rs
ClementBaziret: okay you taught me something then noice

ClementBaziret (approved):
Image

Tholarr (approved):
Image

PR #57, opened 2026-04-05
Basic PPU rendering pipeline

Tholarr:
## 📝 Description

This PR implements a basic PPU rendering pipeline. Each frame, it renders 224 scanlines one by one. For each scanline, it reads tile and color data from two memory regions, VRAM and CGRAM, and outputs the final RGB pixel colors into a framebuffer that is then sent to the display.

VRAM holds two types of data: the tilemap (a grid that describes which tile to draw at each position on screen) and the tile graphics themselves (the actual pixel data for each tile, stored as bitplanes).

CGRAM holds the color palette: up to 256 color entries in BGR555 format, organized into sub-palettes of 16 colors each.

Registers are the control interface: the CPU will writes to them to configure where in VRAM the tilemap and tile data are located, set the scroll position, enable layers, and control brightness.

For each pixel, the renderer walks through the following chain:
```
  Screen pixel (x, y)
        │
        ▼
  Apply BG scroll (BG1HOFS / BG1VOFS)
        │
        ▼
  Tilemap entry (from VRAM)
  ┌─────────────────────────┐
  │ tile index              │
  │ palette number          │
  │ H/V flip flags          │
  └─────────────────────────┘
        │
        ▼
  Tile CHR data (from VRAM)
  4 bitplanes -> 4-bit color index (0..15)
        │
        ▼
  CGRAM lookup
  palette_number * 16 + color_index -> BGR555 color
        │
        ▼
  Brightness scaling + BGR555 -> RGB24
        │
        ▼
  Framebuffer pixel
```

## 🔍 Related Issues

Closes EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier#48
Closes EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier#49
Closes EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier#43


ppu/src/registers.rs
```diff
@@ -0,0 +1,471 @@
...
+    pub slhv: u8, // Software latch for H/V counters
+
+    // $2138 - OAMDATAREAD
+    pub oamdataread: u8, // Read OAM data byte
+
+    // $2139 - VMDATALREAD
+    pub vmdatalread: u8, // VRAM data read low
+
+    // $213A - VMDATAHREAD
+    pub vmdatabread: u8, // VRAM data read high
```
C0Florent: should be `vmdatahread` i assume

ppu/src/vram.rs
```diff
@@ -0,0 +1,303 @@
...
+        self.vmadd = (self.vmadd & 0x7F00) | value as u16;
+        self.vmadd &= 0x7FFF;
+        self.load_latch();
+    }
+
+    pub fn write_vmadd_high(&mut self, value: u8) {
+        self.vmadd = ((value as u16 & 0x7F) << 8) | (self.vmadd & 0x00FF);
+        self.vmadd &= 0x7FFF;
+        self.load_latch();
+    }
```
C0Florent: using u16 helpers from our `common` crate:
```suggestion
    use common::u16_split::U16Split;
    pub fn write_vmadd_low(&mut self, value: u8) {
        *self.vmadd.lo_mut() = value;
        self.load_latch();
    }

    pub fn write_vmadd_high(&mut self, value: u8) {
        *self.vmadd.hi_mut() = value & 0x7F;
        self.load_latch();
    }
```
(of course the `use` should be at the top of the file, not here)

ppu/src/cgram.rs
C0Florent, line 1: Overall looks good, nice and simple, I think it could be simpler with the memory as a `[u16]` instead of a `[u8]`, see #58 for that

ppu/src/renderer.rs
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    fn update_brightness(&mut self, target: u8) {
+        if self.current_brightness != target {
+            if self.brightness_delay == 0 {
+                self.brightness_delay = 72; // 1-chip fade
+            }
+
+            if self.brightness_delay > 0 {
```
C0Florent: `brightness_delay` is always going to be greater than 0 here, did you mean to add an `else` or something?

ppu/src/renderer.rs
```diff
@@ -0,0 +1,163 @@
...
+                self.brightness_delay -= 1;
+
+                if self.current_brightness < target {
+                    self.current_brightness += 1;
+                } else if self.current_brightness > target {
+                    self.current_brightness -= 1;
+                }
+            }
+        }
+    }
```
C0Florent: I think this whole function would read better with early returns, as in replacing
```rs
if a {
 f1();
 if b {
    f2();
    if c {
      f3();
    }
  }
}
```
with
```rs
if !a {
  return;
}
f1();
if !b {
  return;
}
f2();
if !c {
  return;
}
f3();
```

ppu/src/renderer.rs
```diff
@@ -0,0 +1,163 @@
...
+            if force_blank {
+                self.set_pixel(x, y, 0, 0, 0);
+                continue;
+            }
+
+            // Only render in Mode 1 with BG1 enabled
+            if bg_mode != 1 || !bg1_enabled {
+                self.set_pixel(x, y, 0, 0, 0);
+                continue;
+            }
```
C0Florent: I think i'd prefer having these two conditions extracted at the start, then have them early return:

replacing
```rs
for x in [] {
  if a {
    f(x);
    continue;
  }
  if b {
    f(x);
    continue;
  }
  // rest of loop...
}
```
by
```rs
if a || b {
  for x in [] {
    f(x);
  }
  return;
}

for x in [] {
  // rest of loop...
}

File: ppu/src/vram.rs
Comment (C0Florent) [inline ppu/src/vram.rs:1]: similarly to the CGRAM, I think this would be clearer to use a `[u16]` instead of a `[u8]`

this would also help in the renderer, since everything will already be accessible with word addresses (no conversion will be needed)

Comment (C0Florent) [inline ppu/src/vram.rs:1]: The registers implemented here (VMAIN, VMDATA and VMADD) are also in PPURegisters, so they're effectively duplicated, which is probably not good

File: ppu/src/renderer.rs [hunk 6]
```diff
@@ -0,0 +1,163 @@
...
+
+        for x in 0..SCREEN_WIDTH {
+            // Hardware force blank: output black
+            if force_blank {
+                self.set_pixel(x, y, 0, 0, 0);
+                continue;
+            }
+
+            // Only render in Mode 1 with BG1 enabled
+            if bg_mode != 1 || !bg1_enabled {
```
C0Florent: shouldn't this be the exact opposite? this seems confusing: "if we're **not** in mode 1 **or** mode 1 is **not** enabled, then we draw all black"?

to me it would make more sense if it was "if we're in mode 1 **and** mode 1 is **not** enabled, then we draw all black"

ppu/src/renderer.rs
```diff
@@ -0,0 +1,163 @@
...
+                } else if self.current_brightness > target {
+                    self.current_brightness -= 1;
+                }
+            }
+        }
+    }
+
+    pub fn render_scanline(&mut self, ppu: &PPU, y: usize) {
+        let force_blank = ppu.force_blank();
+        let target_brightness = ppu.brightness() & 0x0F;
```
C0Florent: `ppu.brightness()` is already `regs.inidisp & 0x0F`, so no need to `& 0x0F` again here

ppu/src/registers.rs
```diff
@@ -0,0 +1,471 @@
...
+        (self.tm & 0x01) != 0
+    }
+
+    pub fn bg_mode(&self) -> u8 {
+        self.bgmode & 0x07
+    }
+
+    pub fn bg1_tilemap_addr(&self) -> u16 {
+        ((self.bg1sc as u16 & 0xFC) >> 2) * 0x400
+    }
```
C0Florent, line 291: I think the `& 0xFC` is redundant: `0xFC` is `0b11111100`, so `& 0xFC` will only clear the two least significant bits, which are ignored anyways as you right shift by 2 right after (`>> 2`)

basically, assuming you start with `bg1sc` as `0babcdefgh` (where letters are values of individual bits), then `bg1sc` is `0babcdef00` (forcing the two lowest bits to 0), then right shifting by 2 you have `0b00abcdef` (the two lowest bits are shifted out anyways: no need to zero them earlier)

however, since you essentially right shift afterwards by multiplying by `0x400`, the returned value will be `0babcdef0000000000` (10 zeros), which could also be reached by doing `(self.bg1sc as u16 & 0xFC) * 0x100` (clearing the lowest bits with `&` instead of clearing them by right shift.

so i think any of the following would be valid options:
- `(self.bg1sc as u16 >> 2) * 0x400`
- `(self.bg1sc as u16 >> 2) << 10`
- `(self.bg1sc as u16 & 0xFC) * 0x100`
- `(self.bg1sc as u16 & 0xFC) << 8`
- `(self.bg1sc as u16 << 8) & 0xFC00`

ppu/src/registers.rs
```diff
@@ -0,0 +1,471 @@
...
+
+    pub fn bg1_tiledata_addr(&self) -> u16 {
+        (self.bg12nba as u16 & 0x0F) * 0x1000
+    }
```
C0Florent, line 295: From my understanding you could simply return `self.bg12nba as u16 << 12` here, since self.bg12nba is `0bAAAABBBB` where `AAAA` is bg2 stuff, and `BBBB` is what we want to keep (and return as `0bBBBB000000000000` (twelve zeros)), by left shifting by 12 we will completely shift out the `AAAA` bits so we don't even have to worry about them

ppu/src/renderer.rs
```diff
@@ -0,0 +1,163 @@
...
+            }
+
+            // ============================================================
+            // Screen pixel -> tile coordinates
+            // ============================================================
+            let px = (x + scroll_x) & 0xFF;
+            let py = (y + scroll_y) & 0xFF;
+
+            let tile_col = (px / 8) & 31;
+            let tile_row = (py / 8) & 31;
```
C0Florent: useless `& 31` if i'm not mistaken: since `px` and `py` can be at most `0xFF` (255), `px / 8` and `py /  8` will already be at most 31

also you could change `/ 8` to `>> 3` if you find it easier to understand, but i have no preference, either way is fine for me.

ppu/src/renderer.rs
```diff
@@ -0,0 +1,163 @@
...
+            let fine_x = px % 8;
+            let fine_y = py % 8;
```
C0Florent: similar to above, you can change `% 8` to `& 7` (or `& 0b111`) if you find it more readable, but either way works

ppu/src/renderer.rs
```diff
@@ -0,0 +1,163 @@
...
+            let map_byte_addr = map_word_addr * 2;
+
+            let entry_low  = ppu.vram.memory[map_byte_addr];
+            let entry_high = ppu.vram.memory[map_byte_addr + 1];
+            let entry = (entry_low as u16) | ((entry_high as u16) << 8);
+
+            let tile_index = entry & 0x03FF; // bits 9:0
+            let palette_num = (entry >> 10) & 0x07; // bits 12:10
+            let flip_x = (entry & 0x4000) != 0; // bit 14
+            let flip_y = (entry & 0x8000) != 0; // bit 15
```
C0Florent: Bit 13 is unused?

ppu/src/renderer.rs
```diff
@@ -0,0 +1,163 @@
...
+                palette_entry = 0u8
+            } else {
+                // Mode 1 4bpp: palette offset = palette_num * 16
+                palette_entry = (palette_num as u8) * 16 + color_index
+            }
+
+            // ============================================================
+            // Get color from CGRAM and apply brightness
+            // ============================================================
+            let color = ppu.cgram.current_word(palette_entry);
```
C0Florent: weird thing here: `palette_num` is only 3 bits wide, and `color_index` only four (if i understood things correctly), so we only ever reach 128 colours in the CGRAM out of the total 256 colours it can contain.

if we were to take `palette_num = (entry >> 10) & 0x0f` (instead of `0x07`), that would both find a use for the ignored bit 13 of the tilemap entry and allow us to reach all of 256 CGRAM entries

ppu/src/renderer.rs
```diff
@@ -0,0 +1,163 @@
...
+
+            // Brightness scaling: channel = (channel * brightness) >> 4
+            r5 = (r5 * brightness) >> 4;
+            g5 = (g5 * brightness) >> 4;
+            b5 = (b5 * brightness) >> 4;
+
+            // Convert BGR555 -> RGB888
+            let r8 = ((r5.min(31) << 3) | (r5.min(31) >> 2)) as u8;
+            let g8 = ((g5.min(31) << 3) | (g5.min(31) >> 2)) as u8;
+            let b8 = ((b5.min(31) << 3) | (b5.min(31) >> 2)) as u8;
```
C0Florent: I'm wondering: is this right shift by 4 a thing also done in the original SNES? I could imagine the brightness was just applied globally on the screen while keeping the full 15-bit colour "intact"

I'm asking this because here we might be losing some detail (especially in lower brightness) when shifting right which we could preserve a bit more if we don't try to apply the brightness in RGB555 but rather try to combine the RGB555 colour with the brightness to directly get to the RGB888.
In case the original SNES did apply the brighness in RGB555 space then we should definitely do it the same.

ppu/src/renderer.rs
```diff
@@ -0,0 +1,163 @@
+use crate::constants::*;
+use crate::ppu::PPU;
+
+pub struct Renderer {
+    pub framebuffer: Vec<u8>,
```
C0Florent: Could this have a static size instead?
Either `[u8; SCREEN_WIDTH * SCREEN_HEIGHT * 3]` or `Box<[u8; SCREEN_WIDTH * SCREEN_HEIGHT * 3]>`

C0Florent (commented):
Looks like a really good start, most things are correct with potential minor corrections for readability. Registers need to be deduplicated (some are stored in several structs), and i'm a bit unsure about the renderer, since it's the most complex part, it's hard to tell if everything is correct

ppu/src/renderer.rs
```diff
@@ -0,0 +1,163 @@
...
+            let color_index = Self::decode_tile_pixel_from( &ppu.vram.memory, tile_word_base, fx, fy);
+
+            // Color index 0 = transparent => backdrop (palette 0, entry 0)
+            let palette_entry;
+            if color_index == 0 {
+                palette_entry = 0u8
+            } else {
+                // Mode 1 4bpp: palette offset = palette_num * 16
+                palette_entry = (palette_num as u8) * 16 + color_index
+            }
```
C0Florent: Also is this right? this makes us miss 15 potential colours in the CGRAM

Since in the normal case we read from CGRAM at `(palette_num << 4) | color_index`, I'm assuming there are 16 palettes each made of 16 colours, except for some reason the colour at index 0 of all palettes points to the colour 0 of palette 0?
So CGRAM colours at `0x10`, `0x20`, `0x30`, etc till `0xF0` are never read, and simply use the value at `0x00`?
Perhaps these are only used for other rendering modes not implemented yet?

Also you can do `let palette_entry = if a { b } else { c };`

ppu/src/renderer.rs
Tholarr: The fix was applied to `force_blank`.

---

And I've decided to remove the conditions `bg_mode` and `bg1_enabled` for now, since proper handling of them hasn't been fully implemented yet.

The fix will also be applied to them when they are reincluded.

ppu/src/renderer.rs
Tholarr: The `bg_mode` and `bg1_enabled` conditions have been removed for now, as their handling has not yet been fully implemented. Their implementation will be handled differently once they are added back.

ppu/src/registers.rs
Tholarr: The bits are `BBBBAAAA` where `AAAA` (the 4 low bits) is the BG1 CHR base address and `BBBB` (the 4 high bits) is BG2. So `<< 12` would actually shift the BG2 bits out of the u16 entirely, which isn't what we want. The `& 0x0F` mask is intentional here to isolate the BG1 bits before multiplying.

---

[BG12NBA - BG1 and BG2 CHR word base address ($210B write)](https://snes.nesdev.org/wiki/PPU_registers#BG12NBA)
```
7  bit  0
---- ----
BBBB AAAA
|||| ||||
|||| ++++- BG1 CHR word base address (word address = AAAA << 12)
++++------ BG2 CHR word base address (word address = BBBB << 12)
```

Tholarr: After discussing this orally, you were right that `<< 12` works here, the `BBBB` bits naturally overflow out of the u16. I had misread your suggestion as targeting BG2 bits, but the shift achieves the same result as `& 0x0F) * 0x1000` without the explicit mask.

Applying `(self.bg12nba as u16) << 12`.

ppu/src/renderer.rs
Tholarr: According to the [fullsnes documentation](https://problemkaputt.de/fullsnes.htm#snesppucontrol), the brightness formula is `Brightness * (N + 1) / 16`, which means the scaling is indeed applied in RGB555 space before the DAC, so the `>> 4` approach is the hardware-accurate way to do it.

---

But the reviewed implementation was using `channel * N >> 4` instead of `channel * (N + 1) >> 4`, which means at full brightness (N=15) we got 15/16 of the color value instead of 16/16, colors
were never quite at full brightness.
-> So the `N + 1` issue has been fixed.

ppu/src/renderer.rs
Tholarr: That's a good question, but the behavior you're seeing here is correct: the color at index 0 is always transparent.

The previous approach worked, but it wasn't ideal (it was set up for testing and then forgotten). 
-> The issue has been fixed: simply avoid writing to that pixel so as not to overwrite the color it previously had, from previous layers or state.

ppu/src/renderer.rs
Tholarr: Mode 1 intentionally use only the first 128 CGRAM entries. The remaining 128 are used by other layers.

ppu/src/rendering/mode_1.rs
```diff
@@ -0,0 +1,356 @@
...
+    fn decode_4bpp_tile_pixel_from(vram: &[u16], tile_word_base: usize, x: usize, y: usize) -> u8 {
+        // Planes 0+1: words 0-7
+        let w01 = vram[tile_word_base + y];
+        let p0 = (w01 & 0xFF) as u8; // lo byte = plane 0
+        let p1 = (w01 >> 8) as u8; // hi byte = plane 1
+
+        // Planes 2+3: words 8-15
+        let w23 = vram[tile_word_base + 8 + y];
+        let p2 = (w23 & 0xFF) as u8; // lo byte = plane 2
+        let p3 = (w23 >> 8) as u8; // hi byte = plane 3
```
C0Florent: Potentially clearer with `u16::to_le_bytes`
```suggestion
        // Planes 0+1; p1: high byte, p0: low byte
        let [p0, p1] = vram[tile_word_base + y].to_le_bytes();

        // Planes 2+3: words 8-15
        let [p2, p3] = vram[tile_word_base + y + 8].to_le_bytes();
```

see example in playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=6a76db7aa7c9f03cbced5a7f2cdd3c59

ppu/src/vram.rs
C0Florent: deduplicated indeed, but i'm not sure it's in the best way (no idea what the best way would be, i'll have to think on this).
resolving anyways, as the primary problem is solved

ppu/src/rendering/mode_1.rs
```diff
@@ -0,0 +1,356 @@
+use crate::constants::*;
+use crate::ppu::PPU;
+use crate::rendering::renderer::*;
+
+pub trait Mode1Render {
+    fn render_scanline_mode1(&mut self, ppu: &PPU, y: usize);
+    fn decode_4bpp_tile_pixel_from(vram: &[u16], tile_word_base: usize, x: usize, y: usize) -> u8;
+}
```
C0Florent: I'm not sure why we would have a trait for `Mode1Render`, we're probably never going to have multiple implementations of it, right?

One thing we could have more than one implementer of would be a `ModeRender` but even then I'm not sure they would share the same methods, and since anyways we are probably going to `match` against the render mode, I'm not sure we benefit from having a trait at all.

ppu/src/rendering/renderer.rs
```diff
@@ -0,0 +1,329 @@
...
+            framebuffer: Box::new([0; SCREEN_WIDTH * SCREEN_HEIGHT * 3]),
+            current_brightness: 15, // full brightness 
+            brightness_delay: 0,
+        }
+    }
+
+    pub fn render_scanline(&mut self, ppu: &PPU, y: usize) {
+        // Hardware force blank: output black
+        if ppu.force_blank() {
+            Self::render_full_black(self, y);
```
C0Florent: you know that `self.render_full_black(y)` is a thing right?

ppu/src/rendering/mode_1.rs
Tholarr: You are right, the trait was unnecessary since we'll never have multiple implementations of it. I moved the methods directly into an `impl Renderer` block in `mode_1.rs`, to keep the file separation while avoiding the indirection 🫡

C0Florent: Makes sense to me!

ppu/src/rendering/mode_1.rs
```diff
@@ -0,0 +1,348 @@
...
+            let tile_col = px >> 3;
+            let tile_row = py >> 3;
+            let fine_x = px & 7;
+            let fine_y = py & 7;
+
+            // ==========================================================================
+            // Read tilemap entry: tilemap_base is a word address => byte address = * 2
+            // ==========================================================================
+            let map_word_addr = tilemap_base as usize + tile_row * 32 + tile_col;
+            let entry = ppu.vram.memory[map_word_addr];
```
C0Florent, line 32: I think the bit of the comment about the byte address is irrelevant now, since the vram is now using `u16`

C0Florent (dismissed):
👍

tibaudlemaout (dismissed):
lgtm

ppu/src/rendering/renderer.rs
```diff
@@ -0,0 +1,328 @@
...
+
+    pub fn render_scanline(&mut self, ppu: &PPU, y: usize) {
+        // Hardware force blank: output black
+        if ppu.force_blank() {
+            self.render_full_black(y);
+            return;
+        }
+
+        // Update brightness
+        self.update_brightness(ppu.brightness());
```
ClementBaziret, line 28: From what I understand, you update the brightness only once per scanline ?

ppu/src/ppu.rs
```diff
@@ -1,99 +1,828 @@
...
+            // BACKGROUNDS
+            // ==========================
+            0x2105 => self.regs.bgmode = value,
+            0x2106 => self.regs.mosaic = value, // TODO
+            0x2107 => self.regs.bg1sc = value,
+            0x2108 => self.regs.bg2sc = value, // TODO
+            0x2109 => self.regs.bg3sc = value, // TODO
+            0x210A => self.regs.bg4sc = value, // TODO
+            0x210B => self.regs.bg12nba = value, // TODO
+            0x210C => self.regs.bg34nba = value, // TODO
```
ClementBaziret, line 54: Are all those TODO labels because a write to those registers should do something ?

ClementBaziret (dismissed):
This LGTM, I just asked some questions but this should not make the code change so approved 👍.

ppu/src/rendering/mode_1.rs
Tholarr: Good catch !
-> it's fixed 🫡

ppu/src/rendering/renderer.rs
Tholarr: Yes, the same brightness is applied to each pixel of one scanline. Do you think it should be different ?

ppu/src/ppu.rs
Tholarr: The `// TODO` comments are there because the `= value` is a temporary way to “implement” them without actually doing so, and it prevents potential crashes if nothing had been done. So they means that I haven't looked into how to do it yet.

C0Florent (approved):

tibaudlemaout (approved):
lgtm

ppu/src/rendering/renderer.rs
ClementBaziret: Well I don't know, is that the behavior of the SNES PPU ? If so then that's alright 🤷‍♂️.

ppu/src/ppu.rs
ClementBaziret: Okay thanks for the answer.

ppu/src/rendering/renderer.rs
Tholarr: changing the brightness should be "instant", but with the current implementation of the full frame display, it wont change anything to modify it right now. I suggest doing it while making the display accurate (not in this PR)

PR #58, opened 2026-04-07
ppu registers  cgram refac

C0Florent:
Some improvement suggestions for #57, specifically in the CGRAM
- use a custom enum of u8 to keep track of byte phase
- switch from `[u8]` to `[u16]`

I think this overall makes things easier to understand, and it was simpler to just do it than try to explain how to do it

These are just suggestions, feel free to tell me if anything feels wrong to you.

I think similar changes can be done in the VRAM, but i'm not sure yet, i'll have to read it in more detail

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -393,6 +393,7 @@ checksum = "2f3a9f18d041e6d0e102a0a46750538147e5e8992d3b4873aaafee2520b00ce3"
 name = "ppu"
 version = "0.1.0"
 dependencies = [
+ "common",
  "sdl2",
 ]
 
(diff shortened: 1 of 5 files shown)


Tholarr (approved):
Oh, thanks so much for the enum! It was such a pain juggling between the high and low bytes. And the switch from `[u8]` to `[u16]` also seems good to me 👍

C0Florent:
Great, then feel free to merge the PR (potentially "rebase and merge" is more appropriate here), so the changes will just go to your current PPU PR
And also, I definitely think the VRAM should be using a `[u16]` array instead of the current `[u8]`, so if you understood properly how I went about it for the CGRAM, you should be able to do the same for the VRAM yourself (otherwise you can always ask for help)

PR #60, opened 2026-05-01
Handle IO Memory Zone

ClementBaziret:
## 📝 Description

This pull request adds the handling of the many different registers in the IO memory zone (`0x2000` - `0x6000`). There are still some registers that doesn't do anything and that aren't linked, for example all the ppu registers, the interruption registers (RDNMI, NMITIMEN, ...), the APU registers, and some others. However, each of those are written and have `todo!` labels so we know that they aren't implemented.

The main feature added in this pull request is the DMA that has also been implemented. At the moment it doesn't follow how it works in the console itself. The way I implemented it is that it does all the data transfers, then wait the expected time for the DMA to complete in one go with a manual increase of `Rsnes::cpu_master_cycles_to_wait`.

The ideal way would be for it to work kind of like how the cpu works, when a DMA is enabled, it will handle one byte of data, then wait, then another byte and so on and so on. I did not do it like that for now because I was already struggling with implementing the DMA transfer itself. Now that it works, I will later work on making it work like it does in the console originally.

## 🔍 Related Issues

Closes EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier#38

### DIFF
--- a/bus/src/bin/main.rs
+++ b/bus/src/bin/main.rs
@@ -1,12 +1,14 @@
 use bus::bus::Bus;
 use std::{env, error::Error};
 
+#[cfg(not(tarpaulin_include))]
 fn main() {
     if let Err(e) = run() {
         eprintln!("Error: {}", e);
     }
 }
 
+#[cfg(not(tarpaulin_include))]
 fn run() -> Result<(), Box<dyn Error>> {
     let args: Vec<String> = env::args().collect();
     let bus = Bus::new(&args[1])?;
--- a/bus/src/bin/main.rs
+++ b/bus/src/bin/main.rs
@@ -1,14 +1,12 @@
 use bus::bus::Bus;
 use std::{env, error::Error};
 
-#[cfg(not(tarpaulin_include))]
 fn main() {
     if let Err(e) = run() {
         eprintln!("Error: {}", e);
     }
 }
 
-#[cfg(not(tarpaulin_include))]
 fn run() -> Result<(), Box<dyn Error>> {
     let args: Vec<String> = env::args().collect();
     let bus = Bus::new(&args[1])?;
--- a/bus/src/bus.rs
+++ b/bus/src/bus.rs
@@ -1,8 +1,10 @@
 use crate::io::Io;
-use crate::memory_region::MemoryRegion;
 use crate::rom::Rom;
 use crate::wram::Wram;
+use apu::Apu;
 use common::snes_address::SnesAddress;
+use cpu::cpu::CPU;
+use ppu::ppu::PPU;
 use std::error::Error;
 use std::path::Path;
 
@@ -29,11 +31,11 @@ impl Bus {
             [ read ]    [ &self, addr: SnesAddress ]                    [ u8 ]          [ addr ];
             [ write ]   [ &mut self, addr: SnesAddress, value: u8 ]     [ () ]          [ addr, value ];
         ]
-        pub fn DUP_method(DUP_parameters) -> DUP_return_t {
+        pub fn DUP_method(DUP_parameters, cpu: &mut CPU, ppu: &mut PPU, apu: &mut Apu) -> DUP_return_t {
             match addr.bank {
                 0x00..=0x3F | 0x80..=0xBF => match addr.addr {
                     0x0000..0x2000 => self.wram.DUP_method(DUP_method_param),
-                    0x2000..0x6000 => self.io.DUP_method(DUP_method_param),
+                    0x2000..0x6000 => self.io.DUP_method(DUP_method_param, cpu, ppu, apu),
                     0x6000..0x8000 => self.rom.DUP_method(DUP_method_param), // TODO : Expansion port
                     0x8000..=0xFFFF => self.rom.DUP_method(DUP_method_param),
                 },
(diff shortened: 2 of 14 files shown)


bus/src/bin/main.rs
```diff
@@ -1,12 +1,14 @@
 use bus::bus::Bus;
 use std::{env, error::Error};
 
+#[cfg(not(tarpaulin_include))]
```
C0Florent: Don't we already have a rule to ignore all `main.rs` files in the coverage report?
(also applies below)

bus/src/rom/rom.rs
```diff
@@ -136,14 +135,14 @@ impl Rom {
...
-impl MemoryRegion for Rom {
+impl Rom {
     /// Reads a byte from the ROM at the given `SnesAddress`.
     ///
     /// The address is translated to an internal ROM offset using `to_offset`.
     ///
     /// # Panics
     /// Panics if the mapping mode is `MappingMode::Unknown` or index out of bounds.
-    fn read(&self, addr: SnesAddress) -> u8 {
+    pub fn read(&self, addr: SnesAddress) -> u8 {
```
C0Florent, line 144: Should have realised this earlier in a previous review, but the `MappingMode::Unknown` doesn't exist anymore, right (in the doc comment)?

bus/src/bus.rs
```diff
@@ -26,14 +28,14 @@ impl Bus {
     duplicate! {
         [
             DUP_method  DUP_parameters                                  DUP_return_t    DUP_method_param;
-            [ read ]    [ &self, addr: SnesAddress ]                    [ u8 ]          [ addr ];
+            [ read ]    [ &mut self, addr: SnesAddress ]                    [ u8 ]          [ addr ];
             [ write ]   [ &mut self, addr: SnesAddress, value: u8 ]     [ () ]          [ addr, value ];
         ]
```
C0Florent, line 32: unintentionally broke the alignment?

bus/src/io.rs
```diff
@@ -31,39 +140,347 @@ impl Io {
...
+
+            // Joypad data registers
+            0x4218 => self.joy1 as u8,
+            0x4219 => (self.joy1 >> 8) as u8,
+            0x421A => self.joy2 as u8,
+            0x421B => (self.joy2 >> 8) as u8,
+            0x421C => self.joy3 as u8,
+            0x421D => (self.joy3 >> 8) as u8,
+            0x421E => self.joy4 as u8,
+            0x421F => (self.joy4 >> 8) as u8,
```
C0Florent: All of these are *fine*, but might be more readable with `lo()` / `hi()` helpers

bus/src/io.rs
```diff
@@ -13,14 +14,122 @@ use crate::memory_region::MemoryRegion;
 /// For example, the addresses `0x004000` and `0x9E4000` both refer to the
 /// same memory location.
 pub struct Io {
```
C0Florent, line 15: Could we have some basic doc comments on all fields? at least to explain the cryptic names these things have
(same for the DMAChannel below)

bus/src/io.rs
```diff
@@ -13,14 +14,122 @@ use crate::memory_region::MemoryRegion;
...
+    pub joy1: u16,
+    pub joy2: u16,
+    pub joy3: u16,
+    pub joy4: u16,
+
+    pub dma_channels: [DMAChannel; 8],
+}
+
+#[derive(Copy, Clone)]
+pub struct DMAChannel {
```
C0Florent: Why `Copy` on this? it seems weird to me to allow implicit of a 12-byte struct

bus/src/io.rs
```diff
@@ -13,14 +14,122 @@ use crate::memory_region::MemoryRegion;
...
+    pub dash: u8,
+    pub dasb: u8,
+
+    pub a2al: u8,
+    pub a2ah: u8,
+
+    pub nltr: u8,
+
+    pub unused: u8,
+}
```
C0Florent, line 247: Would it make more sense to have some fields be `u16` or `SnesAddress` instead of having `l`/`h` or `l`/`h`/`b` variants?
The specific bytes could still be accessed individually with `u16_split` helpers where needed

Actually this also applies to fields of `Io` above

bus/src/io.rs
```diff
@@ -13,14 +14,122 @@ use crate::memory_region::MemoryRegion;
...
+
```
C0Florent, line 248: Are all the `0xFF` values the actual thing that should be present on init or is it just to make it more obvious that they've not been overwritten yet?

bus/src/io.rs
```diff
@@ -31,39 +140,347 @@ impl Io {
...
+                    _ => {}
+                }
+            }
+
+            _ => {}
+        }
+    }
+
+    #[cfg(not(tarpaulin_include))]
+    fn read_ppu(&mut self, addr: SnesAddress, ppu: &mut PPU) -> u8 {
```
C0Florent, line 510: The ppu read/write functions have already been implemented in https://github.com/r-snes/r-snes/pull/57/changes#diff-b89a95e82fc78f97ed251dff27971450e9a4a888177f18578d92c66f50bd8f21 (except for open bus behaviour) so, should this not be implemented here or?

src/gui.rs
```diff
@@ -15,6 +15,7 @@ pub enum RSnesEvent {
     Quit,
 }
 
+#[cfg(not(tarpaulin_include))]
```
C0Florent, line 18: Would be nice to be able to test gui stuff, but for now it's certainly better to pretend we can't

src/rsnes.rs
```diff
@@ -39,6 +39,96 @@ impl RSnes {
...
+        // Get transfer parameters from channel DMAP register
+        let direction = (ch.dmap >> 7) & 1;
+        let fixed = (ch.dmap >> 3) & 1;
+        let decrement = (ch.dmap >> 4) & 1;
+        let mode = ch.dmap & 0x07;
+
+        let mut a_addr = SnesAddress {
+            bank: ch.a1b,
+            addr: ((ch.a1th as u16) << 8) | ch.a1tl as u16,
+        };
```
C0Florent: this seems to confirm https://github.com/r-snes/r-snes/pull/60/changes#r3203482042: you would definitely benefit from having `a1b`, `a1th` and `a1tl` be a single `SnesAddress`, and likely the same goes for other fields, I'm not going to point out every one of them

src/rsnes.rs
```diff
@@ -39,6 +39,96 @@ impl RSnes {
...
+            1 => &[0, 1],
+            2 | 6 => &[0, 0],
+            3 | 7 => &[0, 0, 1, 1],
+            4 => &[0, 1, 2, 3],
+            5 => &[0, 1, 0, 1],
+            _ => unreachable!(),
+        };
+        let mut pattern_idx = 0;
+
+        while remaining > 0 {
```
C0Florent: Could be `for pattern_idx in 0..remaining {` or something like that? I'm not a fan of having one thing count up and the other count dow, seems like duplicated effort

src/rsnes.rs
```diff
@@ -39,6 +39,96 @@ impl RSnes {
...
+                self.bus
+                    .write(b_addr, byte, &mut self.cpu, &mut self.ppu, &mut self.apu);
+            } else {
+                // B-Bus to A-Bus
+                let byte = self
+                    .bus
+                    .read(b_addr, &mut self.cpu, &mut self.ppu, &mut self.apu);
+                self.bus
+                    .write(a_addr, byte, &mut self.cpu, &mut self.ppu, &mut self.apu);
+            }
```
C0Florent: Could this be `let (src, dest) = if direction == 0 { (a_addr, b_addr) } else { (b_addr, a_addr) };` to deduplicate the read/write calls

C0Florent (commented):
Overall seems good, just a few comments here and there as usual.

I'm starting to think that with more and more things being done in the `bus` crate it might need to be directly integrated in our main crate at some point (to properly time DMA or mul/div register operations).

bus/src/bin/main.rs
ClementBaziret: I don't know about this rule but the coverage is not calculated for the `bin` folder anyway so I removed it 👍 .

bus/src/rom/rom.rs
ClementBaziret: You're right, since the `MappingMode::Unknown` was removed previously, I removed its mention in doc comments.

bus/src/bus.rs
ClementBaziret: Yup, fixed it.

bus/src/io.rs
ClementBaziret: I switched all of the registers that could to either `u16` or `SnesAddress`.

bus/src/io.rs
ClementBaziret: I switched all of the registers that could to either `u16` or `SnesAddress`.

bus/src/io.rs
ClementBaziret: From what I remember when doing it, I made sure to check what the default values were for each register and yeah the DMA registers all init themselves to `0xFF` except `MDMAEN` and `HDMAEN` which are the registers to start the DMA/HDMA which are initialized at 0.

src/gui.rs
ClementBaziret: Yeah this would be great but since this is annoying to try to test and that it will surely change in the future, I think it's fine to leave the `#[cfg(not(tarpaulin_include))]` to ignore for the coverage

src/rsnes.rs
ClementBaziret: I switched all of the registers that could to either `u16` or `SnesAddress`.

bus/src/io.rs
ClementBaziret: `Copy` was on it because of the initialization of the `Io::dma_channels` initialization with `new()`. We changed it with the `Default` attribute and `Copy` and `Clone` aren't needed anymore so I removed them.

src/rsnes.rs
ClementBaziret: Done

src/rsnes.rs
ClementBaziret: Done

bus/src/io.rs
ClementBaziret: I made the modification, I think this should be good now

bus/src/io.rs
ClementBaziret: We can let this function be for now and if the ppu pull request is merged before this one I'll make the change.

bus/src/io.rs
ClementBaziret: I added some documentation on every field so it should now be better.

C0Florent (commented):
Looks all good to me, just remove the cpu -> bus dependency, we'll see later for open bus, perhaps you can have a `todo!` for now, or even an `open_bus()` method which itself is a `todo!`

C0Florent (approved):
:+1:

Tholarr (approved):
lgtm 👍

PR #65, opened 2026-05-16
:memo: Complete CPU documentation

C0Florent:
## 📝 Description

Adds a complete documentation of (almost) all design decisions made to implement the CPU

Also has some minor fixes in the code of CPU, because when reading my code again to write the documentation, I found some things to slightly change


cpu/README.md
```diff
@@ -0,0 +1,733 @@
+# R-SNES CPU software architecture documentation
+
+This is a complete documentation of the relevant decisions made in this crate.
```
ClementBaziret: `relevant decisions` feels a bit weird, maybe `design decisions` is better i don't know.

ClementBaziret (dismissed):

cpu/README.md
C0Florent: Hmm yeah, but I also mention only "relevant" decisions (even though it's kind of obvious)
Perhaps "relevant design decisions" would work better

ClementBaziret: Yep `relevant design decisions` seems good to me 🫡

cpu/docs/README.md
```diff
@@ -0,0 +1,733 @@
...
+This is a complete documentation of the relevant design decisions made in this crate.
+
+<div class="warning markdown-alert markdown-alert-warning"> <!-- we use selectors recognised by github AND rustdoc so it works in both -->
+<!-- We have to use raw HTML links (and other elements) here because rustdoc doesn't parse markdown inside of HTML elements -->
+If you’re just looking for information on how to use this crate, go to the documentation of the <a href="../cpu/struct.CPU.html" title="struct cpu::cpu::CPU"><code>CPU</code></a> struct, or directly to the documentation of <a href="../cpu/struct.CPU.html#method.cycle" title="method cpu::cpu::CPU::cycle"><code>CPU::cycle</code></a> (which is probably all you need to use the CPU in client code)
+</div>
+
+Below are explanations and justifications of the public API of the crate, the internal design of the `cycle` function, and the meta-language used to actually implement instructions. The particular implementations of instructions are not discussed however.
+
+The end of this documentation compares this implementation of the WDC65C816 with implementations found in other emulators.
```
Tholarr, line 12: Maybe adding a clarification so that someone who isn't very familiar with the subject doesn't necessarily have to do research to get an idea of what it's about?
```diff
- [...] the WDC65C816 with [...]
+ [...] the WDC65C816 (a 16-bit microprocessor) with [...]
```

cpu/docs/README.md
```diff
@@ -0,0 +1,733 @@
...
+
+impl CPU {
+    pub fn cycle(&mut self, mem: &mut impl Memory) {
+        // ...
+    }
+}
+```
+(this public API implies that the CPU has arbitrary access the entire addressable space during each cycle, which isn't the case, but we'll come back to this later)
+
+However, in the SNES, CPU cycles may take a different amount of time (counted in master cycles, which take a fixed amount of time) depending on what *type* of cycle the CPU performed. There are three types of cycles: read, write and internal.
```
Tholarr: Why is `type` written in italics?

cpu/docs/README.md
```diff
@@ -0,0 +1,733 @@
+# R-SNES CPU software architecture documentation
+
+This is a complete documentation of the relevant design decisions made in this crate.
+
+<div class="warning markdown-alert markdown-alert-warning"> <!-- we use selectors recognised by github AND rustdoc so it works in both -->
+<!-- We have to use raw HTML links (and other elements) here because rustdoc doesn't parse markdown inside of HTML elements -->
+If you’re just looking for information on how to use this crate, go to the documentation of the <a href="../cpu/struct.CPU.html" title="struct cpu::cpu::CPU"><code>CPU</code></a> struct, or directly to the documentation of <a href="../cpu/struct.CPU.html#method.cycle" title="method cpu::cpu::CPU::cycle"><code>CPU::cycle</code></a> (which is probably all you need to use the CPU in client code)
```
Tholarr: Missing a `.` at the end of the sentence

cpu/docs/README.md
```diff
@@ -0,0 +1,733 @@
...
+    // private fields ...
+}
+
+impl CPU {
+    pub fn cycle(&mut self, mem: &mut impl Memory) {
+        // ...
+    }
+}
+```
+(this public API implies that the CPU has arbitrary access the entire addressable space during each cycle, which isn't the case, but we'll come back to this later)
```
Tholarr, line 30: Missing a `.` at the end of the sentence

cpu/docs/README.md
```diff
@@ -0,0 +1,733 @@
...
+    }
+
+    pub fn addr_bus(&self) -> &SnesAddress {
+        &self.addr_bus
+    }
+}
+```
+
+The idea is that all data transfers go through the public `data_bus` member: when the cycle returns a `Write`, client code using the CPU writes the byte contained in the data bus at the address returned by `addr_bus`, whereas when the CPU returns a read cycle, the client code should write the value in the public `data_bus` field (this is why it is public).
+This way, client code can know where the CPU is accessing memory, and we directly encode the "one byte per cycle" constraint in our API in such a way that it is impossible to break the rule when implementing instruction.
```
Tholarr: Not sure about this one but doesn't `instruction` takes an `s` ?
````diff
- [...] when implementing instruction.
+ [...] when implementing instructions.

File: cpu/docs/README.md [hunk 7]
```diff
@@ -0,0 +1,733 @@
...
+
+        self.next_cycle = next_cycle;
+        ret
+    }
+}
+```
+
+By returning arbitrary function pointers, we are able to account for whatever hardware quirk we may discover we need to implement, of which we found many. The two main categories are:
+
+- Some instructions have "conditional idles" (branch instructions for example), which means they will optionally spend an internal cycle more than usual depending on some condition (for branch instructions: when the condition to branch out is met for example)
```
Tholarr: Missing a `.` at the end of the sentence

cpu/docs/README.md
```diff
@@ -0,0 +1,733 @@
...
+    meta SET_ADDRMODE_IMM;
+    meta FETCH_OP_INTO cpu.registers.A;
+    meta SET_NZ_OP cpu.registers.A;
+});
+```
+
+In a vacuum this seems like a whole lot of magic, but it actually concisely describes all an immediate LDA needs to do:
+
+- `meta SET_OP_SIZE AccMem` set the "variable width" operations to depend on the size of A (which is the case for instructions touching A (the accumulator) or memory, hence the name `AccMem`). This basically "tells" the code generator that in the case it needs to create logical branches based on the size of a variable-width operand, it should use the condition `!cpu.registers.E && !cpu.registers.P.M` (`true` for the 16-bit case).
+- `meta SET_ADDRMODE_IMM` properly sets the address bus to point to an "immediate" operand (an operand located right after the opcode of the instruction)
```
Tholarr: Missing a `.` at the end of the sentence

cpu/docs/README.md
```diff
@@ -0,0 +1,733 @@
...
+    /// Raw function body
+    pub body: TokenStream,
+
+    /// Cycle type (part of the function return value; should evaluate
+    /// to something of type `CycleResult`)
+    pub cyc_type: Ident,
+}
+```
+
+This is really simple: an instruction body is a list of cycles, which all have a fixed cycle type, and a given function body (the rust code which will be the body of the cycle function)
```
Tholarr: Missing a `.` at the end of the sentence

cpu/docs/README.md
```diff
@@ -0,0 +1,733 @@
...
+        }
+        Ok(ret)
+    }
+}
+```
+
+The logic is rather simple, so that it's simple to implement and easily predictable. We start with an empty `InstrBody` (using the `default()` constructor), then in a loop, we (in order):
+
+- Append to the `InstrBody` (in the current cycle we are building up) any tokens found until we encounter a `meta` keyword (which denotes the start of a meta instruction). This is here to allow us to insert any rust code between meta instructions, and it will appear verbatim in the cycle function.
+- (break out of the loop now if there are no more tokens, which will happen if some `InstrBody` ends with raw rust code instead of a meta instruction)
```
Tholarr: Missing a `.` at the end of the sentence

cpu/docs/README.md
```diff
@@ -0,0 +1,733 @@
...
+
+- Just append code in the third cycle<br>
+  This is what some `SET_ADDRMODE_*` meta instrs do (immediate and stack for example, because they don't need cycles to know the address of the operand, so they just set a value in the address bus)
+- Just "close" the third cycle as it is, assigning it a cycle type<br>
+  This is what the `END_CYCLE` meta-instruction does, closing the cycle with a cycle type given as parameter.
+- Append code in the third cycle and close it with a given type<br>
+  This is what the `FETCH8_IMM` meta instr does: it places the address bus to point at an immediate operand, and marks the cycle as `Read`
+- Close the third cycle (giving it a type), and start writing code in the fourth cycle<br>
+  This is what the `FETCH8_IMM_INTO` meta instr does, it first generates the same thing as `FETCH8_IMM` (by calling it directly), and then generates an assignment from the read value to the destination passed as argument in the next cycle
```
Tholarr: For all of the points in this example : 

Missing a `.` at the end of the sentence and `:` before the `<br>` to improve readability and keep the coherence with the last point : `Do the same for any number of cycles: for example append some code in the third, completely generate the fourth and fifth cycles, then start putting some code in the sixth.`

cpu/docs/README.md
```diff
@@ -0,0 +1,733 @@
...
+- Do the same for any number of cycles: for example append some code in the third, completely generate the fourth and fifth cycles, then start putting some code in the sixth.
```
Tholarr, line 309: Do we want to keep the coherence by adding a `<br>` after the `:` ?

cpu/docs/README.md
```diff
@@ -0,0 +1,733 @@
...
+            opcode_fetch(cpu)
+        }
+    }
+}
+```
+
+We could also generate a named function in place of a closure and pass that instead, either way works.
+
+Also, the nice thing with this `InstrBody` type is that we're actually able to use it as the return type of `MetaInstruction::expand()`: meta instrs which just append code to the current cycle but don't create any new cycles can return an `InstrBody` with an empty `cycles` vector but some code in `post_instr`, and meta instrs which want to close a cycle without adding code in it can simply return a `cycles` vector containing a single `Cycle`, which has an empty body, so that when concatenating into another `InstrBody`, it will basically take its `post_instr`, and move it into a new `Cycle` in the `cycles` vector.<br>
+This is what is done in this `AddAssign` impl:
```
Tholarr, line 399: Maybe an other `<br>` at the end of the previous paragraph would improve readability ?

cpu/docs/README.md
```diff
@@ -0,0 +1,733 @@
...
+pub(crate) enum VarWidth<T, U = ()> {
+    /// Same value for 8-bit branch and the 16-bit branch
+    ConstWidth(T),
+
+    /// Potentially different values in the 8- and 16-bit branches
+    /// of the instruction
+    VarWidth{short: T, long: T, data: U},
+}
+```
+(don't pay too much attention to the `data: U` yet, it's almost always `()` anyways)
```
Tholarr, line 519: Missing a `.` at the end of the sentence

cpu/docs/README.md
```diff
@@ -0,0 +1,733 @@
...
+
+Then, we change `InstrBody::parse()` function to return a `VarWidth<InstrBody>`, and meta instr expansions also are `VarWidth<InstrBody>`, so that we can have meta instrs that generate differente code in the 8-bit branch and the 16-bit branch
```
Tholarr: Missing a `.` at the end of the sentence

Tholarr (changes requested):
Just a few minor remarks, but overall it's great 👍

tibaudlemaout (dismissed):
lgtm once all the comments that thomas mentionned are resovled 👍

cpu/docs/README.md
C0Florent: Not sure, I guess it I wanted to stress this word in particular at the time of writing, but now I don't think it's really necessary, it works either way.
Do you think it's better without it? I can't decide which option to choose

cpu/docs/README.md
C0Florent: fixed

cpu/docs/README.md
C0Florent: fixed

cpu/docs/README.md
C0Florent: fixed

cpu/docs/README.md
C0Florent: fixed

cpu/docs/README.md
C0Florent: fixed

cpu/docs/README.md
C0Florent: fixed

cpu/docs/README.md
C0Florent: Fixed (I think)

I intentionally didn't put `.` at the end of the first line of each bullet point item, because they're not actual actual sentences. The exception is the last item which in a sense is the end of a sentence, see example:
To me, the following
> Foo can either:
> - do A
> - do B
> - do C.

reads
> Foo can either do A, do B, or do C.

(the bullet point list replaces linking punctuation, so I'm not adding `.`s which wouldn't be in the full sentence)

---

Then for the last element, where I don't use a line break, I think it makes sense, because in all others, the structure is basically ``Action<br>This is what `meta ...` does.`` (giving a concrete example of a concrete meta-instruction which has the listed behaviour), whereas for the last, the "Action" itself is hard to explain, so my `: for example ...` is explaining the action itself without giving a concrete example of a meta instruction which has this behaviour (perhaps I should add an example to make the formatting more consistent)

Tell me if my changes make sense to you

cpu/docs/README.md
C0Florent: I don't think so, see my comment above

cpu/docs/README.md
C0Florent: Do you mean having two `<br>` one after the other to space them out more? In this case we should just split them in two separate paragraphs.

If that's what you meant I think I agree, It reads a bit better as two separate paragraphs, I'll make the change if that's what you meant.

cpu/docs/README.md
C0Florent: Hmm, actually I'm not sure about that, since it's just a little comment in parentheses, for some reason it makes more sense to me without a period (?)

Is it just me or do you see what I mean?

same answer for https://github.com/r-snes/r-snes/pull/65#discussion_r3343074569

cpu/docs/README.md
C0Florent: fixed

cpu/docs/README.md
C0Florent: I added a description in the top-level README, I think it makes more sense to have it there, does that work in your opinion?

cpu/docs/README.md
C0Florent: https://github.com/r-snes/r-snes/pull/65#discussion_r3343145507

cpu/docs/README.md
Tholarr: Yep that's what I meant :)

cpu/docs/README.md
Tholarr: I didn't see the point of using italics in this context and so from my perspective, I don't see what it adds, so I get the impression there's a meaning I should understand but don't.

cpu/docs/README.md
Tholarr: oh yes, you are right I didn't saw it at the time. It works 👍

cpu/docs/README.md
C0Florent: removed

cpu/docs/README.md
Tholarr: yep, alright 👍

cpu/docs/README.md
C0Florent: fixed

ClementBaziret (approved):

Tholarr (approved):

PR #67, opened 2026-05-30
:sparkles: PPU-bus-CPU linkup

C0Florent:
## 📝 Description

This PR properly links up the CPU, bus and PPU in the main program, so that we're able to run the CPU test ROM with no issues (except for a few missing opcodes which will be implemented in a separate PR)

Depends on #57 (since the entire PPU changed)
Commit 652b7475c0e1 is a very rough implementation of the PPU mode 0, it should be removed once we get a proper PR for it.


bus/src/io.rs
```diff
@@ -409,7 +408,7 @@ impl Io {
         }
     }
 
-    fn write_cpu(&mut self, value: u8, addr: SnesAddress, apu: &mut Apu) {
+    fn write_cpu(&mut self, value: u8, addr: SnesAddress, _: &mut Apu) {
         match addr.addr {
```
ClementBaziret, line 413: I actually left the variable named `apu` on purpose. This function is supposed to use this variable, it is just not at the moment and having the warning at compilation makes a good reminder that it needs to change.
Should we rename it `_` or leave it as is ?

ClementBaziret (commented):
LGTM, just have to use real mode 0 from ppu when it's done and discuss if we want to have real input handling.

bus/src/io.rs
C0Florent: Hmm, fair point, from my point of view I just saw the compiler saying "this is unused", I didn't anticipate you planned to use this as a reminder to use it later.

Reading a few lines below, you already have `todo!("{} : Implement APU channel n°1 reads", addr.addr)`, so hopefully you'll remember that you need to use the `&mut Apu` to implement this, right?

ClementBaziret: Yeah seems fair, it should be fine then 👍

ClementBaziret (approved):

tibaudlemaout (approved):
LGTM and as Clement mentioned it will be revisited when a proper mode 0 is implemented.

PR #68, opened 2026-06-13
cpu interrupts

C0Florent:
## 📝 Description

Implements all software interrupts required to reach the end of the test ROM

Implementation is a bit rough, it relies on the intermediate format produced by the metalang (which is something I try to avoid, to make it easier in case I want to change this intermediate representation, but for now it works, we'll see if/when ever I want to change it)


ClementBaziret (approved):

tibaudlemaout (approved):
Yay interrupts are here.

PR #73, opened 2026-07-04
:sparkles: Link main program <-> plugins <-> CPU

C0Florent:
## 📝 Description

First actual integration of plugins into the main program and CPU module giving access to registers.

For now, the only way to load a plugin is via the CLI option `--load-plugin-noconfirm` which, as it names suggests, does not feature any permission confirmation whatsoever, it grants all permissions requested by the plugin without any user interaction. GUI plugin load and interactive perm confirmation is left to do later.

A nice thing the current CPU integration allows plugins to do is hook a function that will be called at the start of each CPU instruction, receiving as parameter the opcode and current address of execution (as well as all the others CPU registers), which allows plugins to detect what is happening during program execution (although this can be very very laggy, as the lua function will be called about a million times per second).
This is used in the example plugin at cpu/plugin.lua, which is able to count how many tests of the [snes-tests](https://github.com/gilyon/snes-tests) CPU test ROM.

Plugins can be disabled at compile time by disable the `plugins` feature (defined in the root Cargo.toml, and referred to across the main program, to handle whether or not the feature is enabled), to get a version of the program which does not feature any form of plugins (in case this might make some things run faster, or it can result in a smaller executable for example, for people who would like to get a custom-compiled r-snes without this feature)

## 🔍 Related Issues

Closes EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier#68
Closes EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier#70
Closes EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier#63

### DIFF
--- a/.github/workflows/actions.yaml
+++ b/.github/workflows/actions.yaml
@@ -28,7 +28,7 @@ jobs:
         uses: actions/checkout@v5.0.0
 
       - name: Install Rust
-        uses: dtolnay/rust-toolchain@1.92.0
+        uses: dtolnay/rust-toolchain@1.95.0
 
       - name: Install SDL2
         uses: libsdl-org/setup-sdl@main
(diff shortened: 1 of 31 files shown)


Tholarr (approved):
lgtm 👍

tibaudlemaout (approved):
LGTM 👍

ClementBaziret (approved):

PR #75, opened 2026-07-05
Plugins writefile

C0Florent:
## 📝 Description

Allows plugins to write files to disk!

This seemed easy to do initially, but I wanted to allow access of all the various ways a file can be opened for writing in [`OpenOptions`](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html) (truncate, append, create etc.), so this ended up being quite a lot of work.

One thing I struggled with is defining how file write permissions should be ordered (comparing whether some request to write to files is "greater" than another: which is whether it requests "more" permissions, and of course sometimes they can't be compared: neither is greater than the other). Trying to figure out what counts as "more" between opening a file as append without create permissions vs only allowing creating new files but not creating anything vs allow any file creation and overwriting etc. is a lot of cases to think about.
See the doc comments in plugins/src/perm_tree/filesystem.rs for more about that.

And with this, we can now have a plugin which create test reports for snes-test ROMs, saving to a file the number of failed/passed tests, and which tests failed or passed

Depends on #73


ClementBaziret (approved):

tibaudlemaout (approved):
LGTM 👍

PR #80, opened 2026-07-08
Ducktape apu integration

tibaudlemaout:
## 📝 Description

This PR completes the integration of the APU with the rest of the emulator.

The SPC700 now boots through a high-level emulation (HLE) of the IPL boot ROM, communicates with the main CPU through the $2140–$217F communication ports, and successfully runs the entire gilyon/snes-tests SPC700 suite (1,368 tests) through the normal SNES boot process.

The full boot path is now exercised:

1. The main CPU executes spctest.sfc.
2. The test uploads SPC700 code using the standard communication protocol.
3. The SPC700 executes the uploaded code.
4. All 1,368 tests pass.

HLE IPL boot:

The SPC700 now boots through an HLE implementation of the IPL ROM instead of starting directly in user code.

- complete IPL upload protocol ($CC, multi-block uploads, execute handoff)
- correct post-IPL startup state
- support for boot re-entry after jumping back into the IPL region
- timing fixes that eliminate intermittent hangs during multi-block uploads

A skip_ipl_boot() API is also available for tests and tools that initialize ARAM directly.

APU communication

The CPU and APU are now connected through the standard communication ports ($2140–$217F), allowing software to upload and execute SPC700 programs using the normal hardware protocol.

Hardware fixes

Several hardware behaviors were corrected, including:

- proper $F1 (CONTROL) register layout
- implementation of SLEEP and STOP
- various documentation and cleanup improvements within the SPC700 core


src/main.rs
```diff
@@ -15,6 +15,7 @@ use std::{
     time::{Duration, Instant},
 };
 
+//     cargo run --features cli -- /path/to/cputest-basic.sfc
```
Tholarr: Do we want to keep this comment?

tibaudlemaout: dont think it is necessary, i'll remove it.

Tholarr (dismissed):
lgtm

ClementBaziret (dismissed):
LGTM, just did not understand all of the IPL thing but it seems fine to me

Tholarr (dismissed):

apu/tests/fixtures/spc_tests0.spc
C0Florent, line 1: Where do these files come from? If I'm not wrong they're from snes-tests, in which case we should probably link to it and do regular license statements etc.

So I'd add a fixtures/README.md stating that these are from snes-test, and since they also use the MIT license (the same as us for this project) we can also add in the readme that the license for these files is the same as the one found at the root of our repository

apu/src/apu.rs
```diff
@@ -16,35 +84,228 @@ pub struct Apu {
...
+                    self.cpu.regs.a  = 0;
+                    self.cpu.regs.x  = 0;
+                    self.cpu.regs.y  = 0;
+                    self.cpu.regs.pc = entry;
+                    ipl_trace!("[apu ipl] SPC700 running at {entry:#06x}");
+                    self.ipl = IplHle::Done;
+                }
+            }
+
+            IplHle::Done => unreachable!("guarded by caller"),
```
C0Florent: I'm really not a fan of having an `unreachable` for something that has to be checked manually by the caller, this can be very easy to misuse.
Seems slightly tedious to rewrite, so I guess this can be "fine" for now

C0Florent (changes requested):
All good overally, I'd just really like if we could properly source the vendored binary files

apu/tests/fixtures/spc_tests0.spc
tibaudlemaout: they are from the snes test repo that was used to create spctest.sfc, but since the compiled test file has been shared on discord i can now remove them.

apu/src/apu.rs
tibaudlemaout: This has been reworked it so the invariant is in the types instead of the caller: Done is gone, the state is Option<IplHle>, and ipl_step takes the active state and returns the next one, None = handed off. There is no more unreachable.

apu/tests/fixtures/spc_tests0.spc
tibaudlemaout: A README has been added for fixtures.

C0Florent: All good!

apu/src/apu.rs
C0Florent: Great, that works

C0Florent (approved):

Tholarr (approved):
still lgtm

PR #82, opened 2026-07-11
Audible sound during exec

tibaudlemaout:
## 📝 Description

What it does:

With no ROM loaded, the emulator now plays an original looping jingle
after ~3 seconds. The music goes through the entire emulated APU: a
hand-assembled SPC700 driver is uploaded over the real IPL boot protocol
(the frontend acting as main CPU), keeps tempo with hardware timer 0,
and drives the DSP, whose 32 kHz output is queued to the host via SDL2.
Loading a ROM stops it; closing a game brings it back.

Changes

- apu/timers.rs: implement the SPC700 hardware timers (8/8/64 kHz, divisor 0 = 256, 4-bit clear-on-read outputs, enable-edge reset).
- apu/dsp: implement GAIN mode (direct gain + the four ramps), removing the todo!() on register $x7.
- apu/apu.rs: collect one stereo sample pair per DSP tick, drain_samples() for consumers, capped at 1 s if nobody drains.
- apu/jingle.rs (new): mini SPC700 assembler, a 122-byte driver + original melody + looped BRR tone, and the host-side IPL upload, IdleJingle wraps a standalone Apu.
- gui.rs: SDL2 AudioQueue (32 kHz stereo i16) + non-blocking poll_event().
- main.rs: no-ROM branch becomes an idle loop with a 3s delay, boot the jingle, keep ~256 ms queued; failures log once and stay silent.

⚠️ This requires PR#80 to go through first

## 🔍 Related Issues

Closes EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier#84


ClementBaziret (dismissed):

apu/src/apu.rs
```diff
@@ -313,28 +328,44 @@ impl Apu {
...
+
+                // One output sample per DSP tick, straight into the
+                // buffer the host drains. Discard everything if nothing
+                // has drained for a full second (see MAX_BUFFERED_SAMPLES).
+                if self.sample_buf.len() >= MAX_BUFFERED_SAMPLES {
+                    self.sample_buf.clear();
+                }
+                let (l, r) = self.memory.dsp.render_audio_single();
+                self.sample_buf.push(l);
+                self.sample_buf.push(r);
```
C0Florent: If they're always by pairs, why not push the sampled pair directly (having `samble_buf` be a `Vec<(i16, i16)>`)

I think it would also help you since you're reconverting to a `Vec<(i16, i16)>` afterwards anyways

tibaudlemaout: We do indeed always produce them as pairs, but pushing the pair directly would defeat the purpose of the buffer as SDL's audio queue doesn't consume pairs, it takes a `&[i16]`, which is what `drain_samples` hands it. With a `Vec<(i16, i16)>` we'd have to convert and copy it back on every drain.

apu/src/jingle.rs
C0Florent, line 1: Nice addition, almost makes me think long-term we might even want to have a default built-in ROM to boot up when the program starts (since we already have a default framebuffer, and now default audio)

src/main.rs
```diff
@@ -128,6 +128,82 @@ fn gui_emu_loop(
...
+        if let Some(e) = feed_error {
+            eprintln!("waiting music disabled: {e}");
+            gui.audio_stop();
+            jingle = None;
+            jingle_failed = true;
+        }
+
+        // ~256 ms of queued audio makes a 5 ms nap entirely safe, and
+        // keeps this loop invisible in `top`.
+        std::thread::sleep(Duration::from_millis(5));
```
C0Florent: "invisible" is a bit of an overstatement, I'm getting a clean 0% on current main, and fluctuating 3%-5% on your current branch (which makes sense, as we're now actually doing something instead of just waiting for an event truly idle)

tibaudlemaout: the comment has been updated to reflect this.

Tholarr (dismissed):
Lgtm, well done 👍

Tholarr (commented):
Do we want to keep those prints ?

Image

tibaudlemaout:
> Do we want to keep those prints ?
> 
> Image

Given that we haven't been able to properly launch a rom yet and while we think that the missing OAM is the culprit, we dont know how well the cpu<->apu communication holds up with a real ROM i would prefer to keep them at least until we are sure the current IPL is functional.

ClementBaziret (approved):

Tholarr (approved):

PR #83, opened 2026-07-11
:sparkles: Permission request GUI

C0Florent:
## 📝 Description

This PR implements a clean GUI which allows a user to see what permissions a plugin is requesting.
This is not yet linked to the main program, this will have to be linked later.

Since there are no unit tests for this (because I don't even know how I would write UI tests and I don't want to research that now), it would be great if you could try to run the GUI on your machines to see if things look right (`cd plugins; cargo r --features=example`, just be aware that you might get crashes if you press "accept" depending on what the plugin is trying to do, because it is not going to actually receive any of the permissions it requested in this example program).

### DIFF
--- a/plugins/src/bin/main.rs
+++ b/plugins/src/bin/main.rs
@@ -1,3 +1,4 @@
+use egui::ScrollArea;
 use plugins::plugin::{
     Plugin,
     gui::PluginPermRequest,
@@ -21,8 +22,10 @@ impl<'app, 'req> PermRequestApp<'app, 'req> {
 impl eframe::App for PermRequestApp<'_, '_> {
     fn ui(&mut self, ui: &mut egui::Ui, _: &mut eframe::Frame) {
         egui::CentralPanel::default().show_inside(ui, |ui| {
-            ui.heading(&self.title);
-            self.perm_request.show_gui(ui);
+            ScrollArea::both().show(ui, |ui| {
+                ui.heading(&self.title);
+                self.perm_request.show_gui(ui);
+            });
         });
     }
 }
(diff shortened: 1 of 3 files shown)


Tholarr (changes requested):
LGTM, but is it possible to add scroll bar(s) to easily navigate the list of permissions without having to resize the window?

(made with WSL)
Image

C0Florent:
> LGTM, but is it possible to add scroll bar(s) to easily navigate the list of permissions without having to resize the window?

Good call, I'll add that
Would we want the entire window to be scrolling both directions, or maybe have only the perm tree in a resizable scrolling area? I'm also thinking we could have the entire window be a vertical scrolling area, and the perm tree in a horizontal scroller

Tholarr:
> Good call, I'll add that Would we want the entire window to be scrolling both directions, or maybe have only the perm tree in a resizable scrolling area? I'm also thinking we could have the entire window be a vertical scrolling area, and the perm tree in a horizontal scroller

The option to scroll along both axes will be the simplest and fastest to implement. It may not be ideal, but we need to move forward, we can change it later.

C0Florent:
Added scroll to the example @Tholarr
I made the change only in the example (not in the library code), so we will have to use a scrollable in our main program too
This makes more sense in case we want to add content around the current thing (like a title or something), and have the entire thing be scrollable, not just this "component"

ClementBaziret (approved):
LGTM, tested on my side and everything seems fine 👍

tibaudlemaout (approved):
This is great and a really nice preview for what we might end up doing with the plug-ins.

Tholarr (approved):

PR #85, opened 2026-07-12
Fix mapping mode crash

ClementBaziret:
This pull request fixes the problem that was happening when launching game ROM's which consisted of a crash concerning the mapping mode. This should now be fixed.

### DIFF
--- a/bus/src/rom/header/header.rs
+++ b/bus/src/rom/header/header.rs
@@ -50,7 +50,7 @@ impl RomHeader {
         let SpeedAndMappingMode {
             rom_speed,
             mapping_mode,
-        } = SpeedAndMappingMode::from_byte(header_bytes[HEADER_SPEED_MAP_OFFSET]);
+        } = SpeedAndMappingMode::from_byte(header_bytes[HEADER_SPEED_MAP_OFFSET]).unwrap();
 
         RomHeader {
             bytes: header_bytes,
--- a/bus/src/rom/header/mapping_mode.rs
+++ b/bus/src/rom/header/mapping_mode.rs
@@ -218,9 +218,8 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "ERROR: Could not identify mapping of ROM")]
     fn test_from_byte_invalid_mapping_mode() {
-        SpeedAndMappingMode::from_byte(0x02);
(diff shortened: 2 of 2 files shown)


bus/src/rom/header/mapping_mode.rs
```diff
@@ -128,7 +128,7 @@ impl MappingMode {
         let mut score: u32 = 0;
 
         let map_mode = SpeedAndMappingMode::from_byte(rom_data[address + HEADER_SPEED_MAP_OFFSET])
-            .mapping_mode;
+            .map_or(None, |a| Some(a.mapping_mode));
```
C0Florent: `map_or(a, |b| Some(c)` -> `map(|b| c)`

bus/src/rom/header/header.rs
```diff
@@ -50,7 +50,7 @@ impl RomHeader {
         let SpeedAndMappingMode {
             rom_speed,
             mapping_mode,
-        } = SpeedAndMappingMode::from_byte(header_bytes[HEADER_SPEED_MAP_OFFSET]);
+        } = SpeedAndMappingMode::from_byte(header_bytes[HEADER_SPEED_MAP_OFFSET]).unwrap();
```
C0Florent, line 53: Is it really a good idea to unwrap here? Would be great if "bad error handling" was at least only in the main crate, not in component crates

C0Florent (commented):
Good overall, I'm actually surprised this fixes anything because despite changing from a panic to an option, there's a single place where it's actually checked and not unwrapped

bus/src/rom/header/header.rs
ClementBaziret: Yeah I know but if we start to do that I kinda have do to do it for every bad error handling in the bus crate and there is quite a lot sadly.

bus/src/rom/header/mapping_mode.rs
ClementBaziret: Fixed it

bus/src/rom/header/header.rs
C0Florent: Let's say this is good "for now" and we'll come back to it later

Tholarr (approved):

C0Florent (approved):

PR #86, opened 2026-07-12
Sdl2 egui implementation

ClementBaziret:
## 📝 Description

This pull request use the `sdl2-egui` existing crate to use egui with our sdl2 underlying graphic library. It essentially setup egui so that we can add widgets however we want.
While doing that I tested the theory by implementing the "Display ROM Informations" window as a widget, you can try it yourself by pressing F1 during the execution.

Here are some screenshots of what it should look like : 
image
image
image


src/gui/widgets.rs
```diff
@@ -0,0 +1,142 @@
...
+                            ui.label(format!("${:04X}", h.checksum));
+                            ui.end_row();
+
+                            ui.label("Complement");
+                            // The two should XOR to 0xFFFF on a valid cartridge.
+                            let valid = h.checksum ^ h.checksum_complement == 0xFFFF;
+                            ui.label(format!(
+                                "${:04X} {}",
+                                h.checksum_complement,
+                                if valid { "✓" } else { "✗" }
```
C0Florent: These glyphs seem to not be included in the built-in egui fonts as we can see on your screenshots, so perhaps we should use some other symbol or find a way to have the font render them properly?

tibaudlemaout (commented):
This is great, the only thing that could be missing is maybe some documentation about our choice using egui in addition to SDL2 for this feature in particular.

src/gui/state.rs
```diff
@@ -0,0 +1,25 @@
+/// Which overlay windows are currently open.
+///
+/// This lives in `Gui` and persists across frames — egui is immediate-mode,
+/// so the *decision* to draw a window must be stored somewhere that outlives
+/// a single frame. This is that somewhere.
+#[derive(Default)]
+pub struct GuiState {
+    pub show_rom_info: bool,
+    // Future overlays go here:
```
C0Florent: Do we really need this comment? Doesn't hurt much, but we'll probably be able to remember without it?

src/gui/widgets.rs
```diff
@@ -0,0 +1,142 @@
...
+use crate::rsnes::RomInfo;
+
+/// Decodes a SNES header size exponent into KB. `0` means "none".
+fn decode_size_kb(exponent: u8) -> Option<usize> {
+    if exponent == 0 {
+        None
+    } else {
+        Some(1usize << exponent)
+    }
+}
```
C0Florent, line 13: That's actually interesting, I never realised that the size fields stored in the header are exponents and raw sizes. Would it be worth to describe in doc comments in the RomHeader struct?

src/gui/widgets.rs
```diff
@@ -0,0 +1,142 @@
...
+        .vscroll(true)
+        .show(ctx, |ui| {
+            let Some(info) = info else {
+                ui.centered_and_justified(|ui| {
+                    ui.label("No ROM loaded.");
+                });
+                return;
+            };
+
+            let h = &info.header;
```
C0Florent: Similar to what I did in the perm request gui: https://github.com/r-snes/r-snes/blob/a51afdb43cd00a62e7669c3d89b456bf653ef3f1/plugins/src/plugin/gui.rs#L91-L103 Do you think it could be useful to explicitly destructure (without a `..`) so that we can compiler-guarantee we are showing all fields (or if we want to not show a field because it's irrelevant, we manually hide it with `some_irrelevant_field: _` in the destructure.
That way you could also directly format the field names in the curly braces (`format!("{file_size_kb} KB")` insted of `format!("{} KB", info.file_size_kb)`

src/gui.rs
```diff
@@ -49,32 +76,38 @@ impl Gui {
...
         })
     }
 
-    pub fn clear(&mut self, r: u8, g: u8, b: u8) {
-        self.canvas
-            .set_draw_color(sdl2::pixels::Color::RGB(r, g, b));
-        self.canvas.clear();
-    }
+    /// Maps an SDL event to a GUI-only action (toggling overlays).
+    fn map_gui_action(event: &Event) -> Option<GuiAction> {
```
C0Florent: Suggestion: would it make sense to `use sdl2::event::Event as SdlEvent` so that we refer to it as `SdlEvent` (emphasising that it's from sdl directly and not one of our event types, since we have several?)

src/gui.rs
```diff
@@ -132,52 +162,98 @@ impl Gui {
         }
     }
 
-    fn handle_events(&mut self) -> impl Iterator<Item = RSnesEvent> {
-        self.event_pump.poll_iter().filter_map(Self::map_event)
-    }
+    /// Polls SDL, feeds everything to egui, then routes what egui didn't
+    /// consume: first to GUI actions (handled here), then to emulator events
+    /// (returned to the caller).
+    fn handle_events(&mut self) -> Vec<RSnesEvent> {
```
C0Florent, line 185: Slightly disappointing that we now return a `Vec` instead of an iterator, but it's really not that that important, i'll try to change it back some day if i care enough

src/gui.rs
```diff
@@ -132,52 +162,98 @@ impl Gui {
...
         Ok(())
     }
 
+    /// Draws every currently-open overlay. Adding a new window means adding
+    /// one line here plus one field in `GuiState` — nothing in `main.rs`.
+    fn draw_overlays(state: &mut GuiState, data: &GuiFrameData, ctx: &egui::Context) {
+        widgets::rom_info(ctx, &mut state.show_rom_info, data.rom_info.clone());
+    }
+
+    /// One frame: poll input, blit the framebuffer, draw overlays, present.
```
C0Florent, line 254: "blit"? I've never seen that word

src/main.rs
```diff
@@ -81,14 +84,16 @@ fn gui_emu_loop(
         // temporary: toggle VBLANK each rendered frame
         emu_mut.bus.io.rdnmi = !emu_mut.bus.io.rdnmi;
 
-        let events = gui.update(&emu_mut.ppu_renderer.framebuffer);
-        drop(emu_mut); // release the RefCell so that we're able to call plugins
+        let events = gui.update(
+            &emu_mut.ppu_renderer.framebuffer,
+            GuiFrameData {
+                rom_info: Some(rom_info.clone()),
```
C0Florent: We really need to clone the rom_info for each rendered frame? That seems a bit weird, couldn't the gui just use a reference?

Cargo.toml
```diff
@@ -36,6 +36,13 @@ plugins = { version = "0.1.0", path = "./plugins", optional = true }
 piccolo = { workspace = true, optional = true }
 piccolo-util = { workspace = true, optional = true }
 
+egui-sdl2 = { version = "0.4.0", features = ["canvas-backend"] }
+# egui-sdl2 = { version = "0.4.0", features = ["canvas-backend"] }
+# sdl2 and egui are used via the re-exports below, not as direct deps,
+# to guarantee they're the exact same crate instance egui-sdl2 uses:
+#   egui_sdl2::sdl2::*
+#   egui_sdl2::egui::*
+
```
C0Florent: A bit weird that this is here, I've seen no use of `egui_sdl2::sdl2` nor `egui_sdl2::egui` anywhere

TECHNOLOGY_EVALUATION.md
```diff
@@ -65,7 +65,23 @@ Finally, the team explicitly wanted the project to be a vehicle for learning a m
...
+
+Integration required some care around input routing (keeping egui from leaking keyboard/mouse input to the emulated controller) and texture handling, but nothing that changed the overall approach. Framebuffer rendering itself was kept on the SDL2 canvas rather than routed through egui, with egui reserved purely for the floating UI widgets layered on top.
+
+### Decision
+
+**egui-sdl2**, adopted directly on top of the existing SDL2 choice rather than evaluated against alternatives, since it reuses the team's existing egui investment and required no new rendering backend. Scope is intentionally limited to floating debug/introspection widgets; the core framebuffer presentation path remains plain SDL2.
```
C0Florent, line 82: All of that is completely reasonable, I just think it's a bit irrelevant to insist on the fact we want immediate mode UI. Yes, it's true that egui is an immediate-mode UI library (which is one of the things which makes it easy to use) but immediate mode is not fundamentally something we're looking for, we just want some easy-to-use UI library for our needs, and it just so happens that egui is immediate mode

C0Florent (commented):
Very nice progress, just minor changes to do here and there

src/gui/widgets.rs
ClementBaziret: I tried different symbols but not worked or sometimes only the tick would work so I switched to just characters : 
`if valid { "[OK]" } else { "[BAD]" }`.
I feel like this is fine but tell me if you find it ugly, I'll try to figure something else out.

src/gui/state.rs
ClementBaziret: Yep this was to not get lost when adding widgets because there are multiple places where we have to add things for a new widget.
Removed it.

src/gui/widgets.rs
ClementBaziret: I am not sure if this is pertinent, if people want the information, they kind of already know what they are already searching for, I could do it but I am not sure if this is really useful.

src/gui.rs
ClementBaziret: This is used to designate a bit tranfer : https://en.wiktionary.org/wiki/blit

src/main.rs
ClementBaziret: I could have used a reference but from my understanding of the things, I would have to interact with lifetimes and I'm not really comfortable with them. I do not know if clone is a real performance problem, should I switch to a ref ?

Cargo.toml
ClementBaziret: Removed it

src/gui/widgets.rs
C0Florent: As long as it actually renders and is understandable, it's good enough for now

TECHNOLOGY_EVALUATION.md
ClementBaziret: I changed it a bit, tell me if this is better for you or if it needs some other changes.

src/gui.rs
ClementBaziret: Yes it's a shame sorry but I didn't want to bother with it.

src/gui.rs
ClementBaziret: Yep this feels like a good idea, it clears a bit things with all the different "Events" that exists.

ClementBaziret: Done

src/gui/widgets.rs
ClementBaziret: I did not understand what you want to say sorry, could you explain a bit more ?

src/gui/widgets.rs
C0Florent: Ok ok, well it's not that important anyways (feel free to resolve)

src/gui/widgets.rs
C0Florent: The idea is that if you explicitly destructure explicitly as in
```rs
            let RomHeader {
                bytes: _,
                checksum,
                mapping_mode,
                title,
                // and all other fields...
            } = &info.header;
```
you get to explicily list which fields you ignore (for example `bytes: _`: we're not going to render the raw header in the GUI), and have to list all others, because if you forget to list a field, you get a compiler error stating the the pattern is missing fields:
```
error[E0027]: pattern does not mention fields `rom_speed`, `hardware`, `rom_size`, `ram_size`, `country`, `video_standard`, `developer_id`, `rom_version`, `checksum_complement`
  --> src/gui/widgets.rs:32:17
   |
32 |               let RomHeader {
   |  _________________^
33 | |                 bytes: _,
34 | |                 checksum,
35 | |                 mapping_mode,
36 | |                 title,
37 | |             } = &info.header;
   | |_____________^ missing fields `rom_speed`, `hardware`, `rom_size`, `ram_size`, `country`, `video_standard`, `developer_id`, `rom_version`, `checksum_complement`
   ```
   
   This error is usually fixed by adding a `..` pattern at the end to indicate we don't care about other fields, but I think it's actually useful to not use it, this we're forced to list everything, explicitly ignore the things we don't show, and importantly, this serves as a reminder that, if we add a new field to the struct, we have to consider whether or not we want to show in the UI (because otherwise we get a compiler error similar to the one above, and I think that's actually a good thing! at least for the perm tree GUI, it makes a lot of sense to me, perhaps it's also useful for the ROM header GUI?)

File: src/gui.rs (hunk 7 above)
Comment (C0Florent) [reply src/gui.rs:254]: Interesting!

File: TECHNOLOGY_EVALUATION.md (hunk 10 above)
Comment (C0Florent) [reply TECHNOLOGY_EVALUATION.md:82]: Looks alright to me

File: src/main.rs (hunk 8 above)
Comment (C0Florent) [reply src/main.rs]: The only slightly annoying part is storing a reference inside a struct, in which case the struct has to become generic over the lifetime of the stored reference.
This seems to do the trick, and should result in 0 cloning
```diff
diff --git a/src/gui.rs b/src/gui.rs
index 27b8769..e277ca1 100644
--- a/src/gui.rs
+++ b/src/gui.rs
@@ -29,8 +29,8 @@ pub struct Gui {
 /// Add a field here when a new overlay needs new data; `main.rs` never
 /// touches egui directly.
 #[derive(Default)]
-pub struct GuiFrameData {
-    pub rom_info: Option<RomInfo>,
+pub struct GuiFrameData<'a> {
+    pub rom_info: Option<&'a RomInfo>,
 }
 
 #[derive(PartialEq, Eq, Debug)]
@@ -248,7 +248,7 @@ impl Gui {
     /// Draws every currently-open overlay. Adding a new window means adding
     /// one line here plus one field in `GuiState` — nothing in `main.rs`.
     fn draw_overlays(state: &mut GuiState, data: &GuiFrameData, ctx: &egui::Context) {
-        widgets::rom_info(ctx, &mut state.show_rom_info, data.rom_info.clone());
+        widgets::rom_info(ctx, &mut state.show_rom_info, data.rom_info);
     }
 
     /// One frame: poll input, blit the framebuffer, draw overlays, present.
diff --git a/src/gui/widgets.rs b/src/gui/widgets.rs
index 38ffa91..4a42664 100644
--- a/src/gui/widgets.rs
+++ b/src/gui/widgets.rs
@@ -12,7 +12,7 @@ fn decode_size_kb(exponent: u8) -> Option<usize> {
 }
 
 /// ROM information overlay. Toggled with F1.
-pub fn rom_info(ctx: &egui::Context, open: &mut bool, info: Option<RomInfo>) {
+pub fn rom_info(ctx: &egui::Context, open: &mut bool, info: Option<&RomInfo>) {
     egui::Window::new("ROM Information")
         .open(open)
         .resizable(true)
diff --git a/src/main.rs b/src/main.rs
index cbfa578..f66d95d 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -89,7 +89,7 @@ fn gui_emu_loop(
         let events = gui.update(
             &emu_mut.ppu_renderer.framebuffer,
             GuiFrameData {
-                rom_info: Some(rom_info.clone()),
+                rom_info: Some(&rom_info),
             },
         );
         drop(emu_mut);
```
I didn't even need to change the case when no ROM is loaded, `GuiFrameData::default()` handles it already.
There's also no complication since this `GuiFrameData` is constructed each frame, so the borrow isn't long-lasting, it's created just for the render time and dropped after so that it's available for the emulator to use

ClementBaziret: Okay, fixed it with your changes 🫡

src/gui/widgets.rs
ClementBaziret: Okay I understand now thanks for the explanation. I made the modification, tell me if it corresponds to what you had in mind.

C0Florent: Yup, great, just as a bonus now you can also shorten the calls to `format!` if you prefer it that way (previously wasn't possible because the values were in `h.somefield`, but now since they're just `somefield` you can format them with `format!("someformat {somefield}")`

ClementBaziret: Done the small modification

tibaudlemaout (approved):

C0Florent (approved):

PR #88, opened 2026-07-14
:sparkles: Show runtime errors in GUI instead of panic!ing

C0Florent:
## 📝 Description

Previously, the program would just crash as soon as any error occured (invalid ROM, plugin error, etc.)
Now that we have egui to render things, we can at least catch the error and display it to the user (not in the best way but that's still way better than hard crashing)

## 🔍 Related Issues

Closes EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier#<issue-number>

### DIFF
--- a/demo_minimal.lua
+++ b/demo_minimal.lua
@@ -1,0 +1,6 @@
+return {
+    permissions = "none",
+    init = function()
+        print("Hello world")
+    end,
+}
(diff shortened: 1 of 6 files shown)


tibaudlemaout (approved):
LGTM 👍

Tholarr (approved):
verinice 👍

PR #90, opened 2026-07-15
:sparkles: Bus plugin module

C0Florent:
## 📝 Description

Adds a plugin module which allows arbitrary read/writes through the bus, allowing reads and writes anywhere in memory, and IO registers etc.


Tholarr (approved):
lgtm

tibaudlemaout (approved):

PR #91, opened 2026-07-15
:sparkles: PPU plugin module

C0Florent:
## 📝 Description

Minimal plugin PPU module, which has a single function: `write_cgram`, which writes a u16 in the PPU CGRAM.

An example plugin uses this module to change the colour of the snes-tests ROMS


Tholarr (approved):
lgtm

tibaudlemaout (approved):

PR #98, opened 2026-08-17
PPU OAM

Tholarr:
## 📝 Description

This PR implements the OAM (Object Attribute Memory) subsystem and sprite rendering for the PPU.

What's included:

- OAM struct with register access (word-based addressing, write-twice for the low table, immediate writes for the high table)
- Sprite decoding: position, tile, palette, priority, flips, and size
- Per-scanline evaluation with the 32-sprite limit, priority rotation, and STAT77 time-over / range-over flags
- Per-pixel priority buffer for correct z-ordering between backgrounds and sprites
- Sprite rendering: multi-tile sprites (8x8 up to 64x64), flips, sprite palettes, transparency
- Unit tests and a visual example in main (cf. image)

Visual example: BG1 (gradient background) with an 8x8 red sprite and a 16x16 green sprite rendered on top
image

## 🔍 Related Issues

Closes EpitechPromo2027/G-EIP-600-NAN-6-1-eip-florent.charpentier#85

### DIFF
--- a/ppu/src/oam.rs
+++ b/ppu/src/oam.rs
@@ -321,6 +321,9 @@ impl OAM {
 mod tests {
     use super::*;
 
+    // (objsel, small size, large size) for sprite_size cases
+    type SizeCase = (u8, (u8, u8), (u8, u8));
+
     fn make_oam() -> OAM {
         let mut oam = OAM::new();
         // Park every sprite off-screen (y=240) by default, so evaluation
@@ -451,7 +454,7 @@ mod tests {
     // sprite_size returns correct dimensions for all 6 documented size modes.
     #[test]
     fn test_sprite_size() {
-        let cases: &[(u8, (u8, u8), (u8, u8))] = &[
+        let cases: &[SizeCase] = &[
             (0 << 5, (8, 8), (16, 16)),
             (1 << 5, (8, 8), (32, 32)),
             (2 << 5, (8, 8), (64, 64)),
--- a/ppu/src/oam.rs
+++ b/ppu/src/oam.rs
@@ -5,6 +5,33 @@ const OAM_TABLE1_SIZE: usize = 512;
 const OAM_TABLE2_SIZE: usize = 32;
 const OAM_SIZE: usize = OAM_TABLE1_SIZE + OAM_TABLE2_SIZE;
 
+/// Decoded sprite attributes from OAM.
+#[derive(Debug, Clone, Copy)]
+pub struct Sprite {
+    /// X position (9-bit signed: sprites past 255 wrap to the left of the screen)
+    pub x: i16,
+    /// Y position (top edge of the sprite)
+    pub y: u8,
+    /// Raw tile number (0-255) from OAM. The final VRAM address is
+    /// `chr_base + tile * 16`, with multi-tile sprites wrapping the low nibble
+    /// (horizontal) and high nibble (vertical) independently.
+    pub tile: u8,
+    /// Base CHR word address for this sprite, with name-table selection applied.
+    pub chr_base: u16,
+    /// Name-table select bit (from table 1 attribute byte)
+    pub name_table: u8,
+    /// Palette number (0-7); sprites use CGRAM entries 128-255
+    pub palette: u8,
+    /// Priority (0-3)
+    pub priority: u8,
+    /// Horizontal flip
+    pub flip_x: bool,
+    /// Vertical flip
+    pub flip_y: bool,
+    /// Large size bit (the actual size comes from OBJSEL)
+    pub large: bool,
+}
+
 pub struct OAM {
     /// Raw OAM data: 512 bytes table 1 + 32 bytes table 2.
     data: [u8; OAM_SIZE],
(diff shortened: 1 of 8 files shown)


ppu/src/rendering/renderer.rs
```diff
@@ -4,10 +4,33 @@ use crate::ppu::PPU;
...
+// Only BG1 and OBJ are rendered for now; the values are spaced so the other
+// layers slot in later without renumbering.
+// ============================================================
+pub const Z_BACKDROP: u8 = 0;
+pub const Z_OBJ0: u8 = 3;
+pub const Z_OBJ1: u8 = 6;
+pub const Z_BG1_LOW: u8 = 8;
+pub const Z_OBJ2: u8 = 9;
+pub const Z_BG1_HIGH: u8 = 11;
+pub const Z_OBJ3: u8 = 12;
```
C0Florent: Wouldn't this work better as an enum? That way you can write all your variants in order without having to manully assign an arbitrary number for sorting, and put [`#[derive(Ord, PartialOrd)]`](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#derivable) on the enum so that you can still use `>`, `<=` and such.

If this does become an enum, then you might as well [`#[derive(Default)]` and put `#[default]`](https://doc.rust-lang.org/std/default/trait.Default.html#enums) on the "backdrop" variant, it seems to make sense as the "default" value

ppu/src/oam.rs
```diff
@@ -0,0 +1,596 @@
...
+    pub flip_x: bool,
+    /// Vertical flip
+    pub flip_y: bool,
+    /// Large size bit (the actual size comes from OBJSEL)
+    pub large: bool,
+}
+
+pub struct OAM {
+    /// Raw OAM data: 512 bytes table 1 + 32 bytes table 2.
+    data: [u8; OAM_SIZE],
```
C0Florent: What do you think of splitting the two tables in two separate fields?
Since you already have to check which table we're addressing in each read/write (because of the high table being mirrored 8 times), I think it would make sense to have them as two separate fields

ppu/src/oam.rs
```diff
@@ -0,0 +1,596 @@
...
+    ///   stored here (this borrows &self); call `set_flags` to latch them.
+    pub fn eval_sprites_for_scanline(
+        &self,
+        y: usize,
+        objsel: u8,
+        oamadd: u16,
+    ) -> (Vec<(u8, Sprite)>, bool, bool) {
+        let priority_rotation = (oamadd >> 8) & 0x01 != 0;
+        let start = if priority_rotation {
+            ((oamadd & 0xFF) >> 1) as usize & 0x7F
```
C0Florent: Useless `&0x7F`: `oamadd & 0xFF` is at most 8 bits, then after the right shift `(oamadd & 0xFF) >> 1` is at most 7 bits already

C0Florent: Also might as well replace `oamadd & 0xFF` by `*oamadd.lo()`

ppu/src/oam.rs
```diff
@@ -0,0 +1,596 @@
...
+    ///   that, `range_over` is set.
+    /// - Returns (visible sprites, time_over, range_over). The flags are not
+    ///   stored here (this borrows &self); call `set_flags` to latch them.
+    pub fn eval_sprites_for_scanline(
+        &self,
+        y: usize,
+        objsel: u8,
+        oamadd: u16,
+    ) -> (Vec<(u8, Sprite)>, bool, bool) {
+        let priority_rotation = (oamadd >> 8) & 0x01 != 0;
```
C0Florent: I'm pretty sure this `priority_rotation` extraction from `oamadd` is wrong: if I understand correctly [this](https://snes.nesdev.org/wiki/PPU_registers#OAMADDL,_OAMADDH_-_OAM_word_address_($2102,_$2103_write)), `priority_rotation` refers to the bit noted P, which should be accessed by `oamadd >> 15 != 0`. I could be misunderstanding the thing, so try to double check yourself too.

I then tried to investigate:

- Replacing the `(oamadd >> 8) & 0x01 != 0` by either `true` or `false` doesn't make any unit test fail, so this currently untested
- I went in the unit test `test_eval_priority_rotation` below (seems like the one which is meant to test this), and to my understand it indeed doesn't test this at all. The only visible difference between this bit being set or clear is the order in which the visible sprites will be pushed in the returned `Vec<(u8, Sprite)>`, but the test only adds a single visible sprite, so there is no order to compare.
- My fix for the test would be something like this (using two sprites so that we have an order to compare, and also test both with priority_rotation on and off):
```diff
diff --git a/ppu/src/oam.rs b/ppu/src/oam.rs
index 970862f..8551949 100644
--- a/ppu/src/oam.rs
+++ b/ppu/src/oam.rs
@@ -520,11 +520,22 @@ mod tests {
     #[test]
     fn test_eval_priority_rotation() {
         let mut oam = make_oam();
+        make_sprite_entry(&mut oam, 00, 0, 0, 0, 0, 0);
         make_sprite_entry(&mut oam, 10, 0, 0, 0, 0, 0);
-        let oamadd: u16 = (1 << 8) | 20; // enable + start sprite 10 (20 >> 1)
+
+        let oamadd: u16 = (0 << 8) | 20; // disable priority rotation + start sprite 10 (20 >> 1)
         let (visible, _, _) = oam.eval_sprites_for_scanline(0, OBJSEL_8_16, oamadd);
-        assert_eq!(visible.len(), 1);
+        // we expect 0 to come before 10, because we start from 0
+        assert_eq!(visible.len(), 2);
+        assert_eq!(visible[0].0, 0);
+        assert_eq!(visible[1].0, 10);
+
+        let oamadd: u16 = (1 << 8) | 20; // enable priority rotation + start sprite 10 (20 >> 1)
+        let (visible, _, _) = oam.eval_sprites_for_scanline(0, OBJSEL_8_16, oamadd);
+        // we expect 10 to come before 0, because we start at 10 and only reach 0 after wrap around
+        assert_eq!(visible.len(), 2);
         assert_eq!(visible[0].0, 10);
+        assert_eq!(visible[1].0, 0);
     }
 
     // Y wraps in u8: a sprite at y=250 covers scanline 255 but not 249.
  ```
- But then even that test still passes because the test too is using a shift of 8 (in `1 << 8`) to pass the priority_rotation bit.
  So of course, if I change the test to left shift 15 instead of 8, then it doesn't pass, because we would have to right shift 15 in the implementation, but this isn't a proof that the correct value is 8 or 15.

So my question is: Do you agree this should be 15 instead of 8?

File: ppu/src/sprites.rs [hunk 5]
```diff
@@ -0,0 +1,79 @@
...
+                // Locate the tile inside the sprite and the pixel inside the tile.
+                let tile_col = sx / 8;
+                let tile_row = sy / 8;
+                let fine_x = sx % 8;
+                let fine_y = sy % 8;
+
+                // Multi-tile sprites wrap the low nibble (X) and high nibble (Y) of the tile number independently.
+                let base = sprite.tile as usize;
+                let tx = (base & 0x0F).wrapping_add(tile_col) & 0x0F;
+                let ty = (base & 0xF0).wrapping_add(tile_row * 0x10) & 0xF0;
```
C0Florent: I get the intent behing using `wrapping_add`, replicate hardware behaviour which wraps around...
But here the `wrapping_add` is done on a `usize` so there's no chance it ever gets to wrap when adding numbers which fit in `u8`/`u16`, you would actually still get numbers above 255 when doing `(base & 0xF0).wrapping_add(tile_row * 0x10)` if `base` and `tile_row` are big enough.

Since you `& 0x0F` and `& 0xF0` at the end, you're all good at least, that wraps the values back in 0-255.

To remove the redundant `wrapping_add`, you can either just use `+` on `usize` or do `wrapping_add` on `u8`s

C0Florent (changes requested):
Good work overall, seems like it has a chance of working first try when we try to run actual games (judging from the few bits i compared from documentation)

Most comments are just refactor ideas I had while reading the code, feel free to either take them or say "we'll see later" if you feel like i'm nitpicking too much

(Requesting changes because one detail I think is an actual mistake)

ppu/src/rendering/renderer.rs
Tholarr: Good idea, I like it. Since the spaced-out constants currently exist so I can slot in the missing layers (BG2/BG3/BG4 priorities) later, it can make sense to switch to an enum at the same time as I add those remaining variants. That way I do the reordering once, with everything in place. If that works for you, I'll open an issue to track it and integrate it then rather than in this PR.

ppu/src/oam.rs
Tholarr: I see the point, especially for `get_sprite`. But the word-based addressing sees the OAM as one continuous space of 544 bytes, so `read_data`/`write_data `still need to check "which table am I in" anyway. So I don't think splitting the fields would really help there.

ppu/src/oam.rs
Tholarr: done 🫡

ppu/src/oam.rs
Tholarr: fixed 🫡

ppu/src/sprites.rs
Tholarr: done 🫡

ppu/src/rendering/renderer.rs
Tholarr: Just to follow up on this: we talked about it in person and I agreed to make the change, but thinking it over I'd rather hold it for a later PR. I'd do the switch together with the missing layer priorities (BG2/BG3/BG4), otherwise I'd reorder now and again later. And since it's a readability/maintainability change with no impact on behaviour (the current constants produce the exact same z-order), holding it doesn't block anything here. I'll track it in an issue so it doesn't get lost.

ppu/src/oam.rs
C0Florent: good! thanks for adding the test too!

ppu/src/rendering/renderer.rs
C0Florent: Fine, you can do in the next PR instead if that works better for you
I just tried to see what it could look like and got this which compiles, use it if/when you need it

```diff
diff --git a/ppu/src/rendering/mode_0.rs b/ppu/src/rendering/mode_0.rs
index 5f6be90..70558c1 100644
--- a/ppu/src/rendering/mode_0.rs
+++ b/ppu/src/rendering/mode_0.rs
@@ -1,6 +1,6 @@
 use crate::constants::*;
 use crate::ppu::PPU;
-use crate::rendering::renderer::{Renderer, Z_BG1_HIGH, Z_BG1_LOW};
+use crate::rendering::renderer::{RenderPriority, Renderer};
 use crate::vram::RawVRAM;
 
 impl Renderer {
@@ -57,7 +57,11 @@ impl Renderer {
             let color = ppu.cgram.read(palette_entry);
 
             let (r, g, b) = Self::apply_brightness(color, self.current_brightness as u16);
-            let z = if priority { Z_BG1_HIGH } else { Z_BG1_LOW };
+            let z = if priority {
+                RenderPriority::Bg1High
+            } else {
+                RenderPriority::Bg1Low
+            };
             self.set_pixel_z(x, y, r, g, b, z);
         }
     }
diff --git a/ppu/src/rendering/mode_1.rs b/ppu/src/rendering/mode_1.rs
index 3939699..7731e6d 100644
--- a/ppu/src/rendering/mode_1.rs
+++ b/ppu/src/rendering/mode_1.rs
@@ -1,6 +1,6 @@
 use crate::constants::*;
 use crate::ppu::PPU;
-use crate::rendering::renderer::{Renderer, Z_BG1_HIGH, Z_BG1_LOW};
+use crate::rendering::renderer::{RenderPriority, Renderer};
 use crate::vram::RawVRAM;
 
 impl Renderer {
@@ -57,7 +57,11 @@ impl Renderer {
             let color = ppu.cgram.read(palette_entry);
 
             let (r, g, b) = Self::apply_brightness(color, self.current_brightness as u16);
-            let z = if priority { Z_BG1_HIGH } else { Z_BG1_LOW };
+            let z = if priority {
+                RenderPriority::Bg1High
+            } else {
+                RenderPriority::Bg1Low
+            };
             self.set_pixel_z(x, y, r, g, b, z);
         }
     }
diff --git a/ppu/src/rendering/renderer.rs b/ppu/src/rendering/renderer.rs
index eae36c8..8036a10 100644
--- a/ppu/src/rendering/renderer.rs
+++ b/ppu/src/rendering/renderer.rs
@@ -4,24 +4,35 @@ use crate::ppu::PPU;
 /// Raw byte array of the framebuffer in RGB888
 pub type RawFramebuffer = [u8; SCREEN_WIDTH * SCREEN_HEIGHT * 3];
 
-// ============================================================
-// Z-order (priority) values, mode 1 (BG1 + OBJ subset).
-// Higher = closer to the front. A pixel is only overwritten when the
-// incoming z is >= the z already stored for that column.
-//
-// Full mode-1 order (front to back), for reference:
-//   OBJ.3 > BG1.1 > BG2.1 > OBJ.2 > BG1.0 > BG2.0 > OBJ.1 >
-//   BG3.1 > BG4.1 > OBJ.0 > BG3.0 > BG4.0 > backdrop
-// Only BG1 and OBJ are rendered for now; the values are spaced so the other
-// layers slot in later without renumbering.
-// ============================================================
-pub const Z_BACKDROP: u8 = 0;
-pub const Z_OBJ0: u8 = 3;
-pub const Z_OBJ1: u8 = 6;
-pub const Z_BG1_LOW: u8 = 8;
-pub const Z_OBJ2: u8 = 9;
-pub const Z_BG1_HIGH: u8 = 11;
-pub const Z_OBJ3: u8 = 12;
+/// # Z-order (priority) values, mode 1 (BG1 + OBJ subset).
+/// Higher = closer to the front. A pixel is only overwritten when the
+/// incoming z is >= the z already stored for that column.
+///
+/// Full mode-1 order (front to back), for reference:
+///   OBJ.3 > BG1.1 > BG2.1 > OBJ.2 > BG1.0 > BG2.0 > OBJ.1 >
+///   BG3.1 > BG4.1 > OBJ.0 > BG3.0 > BG4.0 > backdrop
+/// Only BG1 and OBJ are rendered for now.
+///
+/// ---
+///
+/// The PartialOrd derive defines comparison operators to reflect
+/// the order of declaration, so since Bg1Low is declared before
+/// Bg1High, we have:
+/// ```
+/// # use ppu::rendering::renderer::RenderPriority;
+/// assert!(RenderPriority::Bg1Low < RenderPriority::Bg1High);
+/// ```
+#[repr(u8)]
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
+pub enum RenderPriority {
+    Backdrop,
+    Obj0,
+    Obj1,
+    Bg1Low,
+    Obj2,
+    Bg1High,
+    Obj3,
+}
 
 pub struct Renderer {
     pub framebuffer: Box<RawFramebuffer>,
@@ -29,7 +40,7 @@ pub struct Renderer {
 
     /// Per-column z-order of the pixel currently written on the scanline
     /// being rendered. Reset to Z_BACKDROP at the start of each scanline.
-    priority: Box<[u8; SCREEN_WIDTH]>,
+    priority: Box<[RenderPriority; SCREEN_WIDTH]>,
 
     brightness_delay: u8,
 }
@@ -45,7 +56,7 @@ impl Renderer {
         Self {
             framebuffer: Box::new([0; SCREEN_WIDTH * SCREEN_HEIGHT * 3]),
             current_brightness: 15, // full brightness
-            priority: Box::new([Z_BACKDROP; SCREEN_WIDTH]),
+            priority: Box::new([RenderPriority::Backdrop; SCREEN_WIDTH]),
             brightness_delay: 0,
         }
     }
@@ -61,7 +72,7 @@ impl Renderer {
         self.update_brightness(ppu.brightness());
 
         // Reset priorities and fill with the backdrop
-        self.priority.fill(Z_BACKDROP);
+        self.priority.fill(RenderPriority::Backdrop);
         let backdrop = ppu.cgram.read(0);
         let (br, bg, bb) = Self::apply_brightness(backdrop, self.current_brightness as u16);
         for x in 0..SCREEN_WIDTH {
@@ -128,7 +139,7 @@ impl Renderer {
 
     /// Write a pixel only if its z-order is at least the one already stored
     /// for this column. On success, updates the stored z-order.
-    pub fn set_pixel_z(&mut self, x: usize, y: usize, r: u8, g: u8, b: u8, z: u8) {
+    pub fn set_pixel_z(&mut self, x: usize, y: usize, r: u8, g: u8, b: u8, z: RenderPriority) {
         if z >= self.priority[x] {
             self.set_pixel(x, y, r, g, b);
             self.priority[x] = z;
diff --git a/ppu/src/sprites.rs b/ppu/src/sprites.rs
index cae194a..f19635c 100644
--- a/ppu/src/sprites.rs
+++ b/ppu/src/sprites.rs
@@ -1,7 +1,7 @@
 use crate::constants::*;
 use crate::oam::OAM;
 use crate::ppu::PPU;
-use crate::rendering::renderer::{Renderer, Z_OBJ0, Z_OBJ1, Z_OBJ2, Z_OBJ3};
+use crate::rendering::renderer::{RenderPriority, Renderer};
 
 // VRAM is 32768 words, sprite CHR addresses wrap within it.
 const VRAM_WORD_MASK: usize = (VRAM_SIZE / 2) - 1;
@@ -28,10 +28,10 @@ impl Renderer {
             }
 
             let z = match sprite.priority {
-                0 => Z_OBJ0,
-                1 => Z_OBJ1,
-                2 => Z_OBJ2,
-                _ => Z_OBJ3,
+                0 => RenderPriority::Obj0,
+                1 => RenderPriority::Obj1,
+                2 => RenderPriority::Obj2,
+                _ => RenderPriority::Obj3,
             };
 
             for col in 0..w {
```

ppu/src/oam.rs
C0Florent: :shrug:

C0Florent (approved):
:+1:

ClementBaziret (approved):

